PR #12, opened 2018-10-11
implement --format-json JSON option. It's usefull for post-processing…

lebouquetin:
This implementation limit dependencies as `json` dependency will not be required except if you try to export as JSON. In order to allow this, the `import json` has been moved to the unique method requiring json : `JsonPrettyTable.get_string()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ Dump the software license list of Python packages installed with pip.
     * [Option: format\-rst](#option-format-rst)
     * [Option: format\-confluence](#option-format-confluence)
     * [Option: format\-html](#option-format-html)
+    * [Option: format\-json](#option-format-json)
     * [More Information](#more-information)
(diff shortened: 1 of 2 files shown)


lebouquetin:
@raimon49 the CI fails for python 3.3 for an unknown reason - not related to the current MR. I may help if requested, but I really don't know the reason for this failure

raimon49:
@lebouquetin Thanks for the good JSON support. 

It fails in Python 3.3, probably because it dropped 3.3 support in pip. pip-licenses will also drop support for Python 3.3 in the near future.

I am planning to merge your patch.

raimon49:
@lebouquetin Released version 1.8.0 with your added option enabled.
https://pypi.org/project/pip-licenses/1.8.0/

Thanks for your contribution.

PR #14, opened 2018-11-15
Adds a summary argument

jnhmcknight:
This allows the user to pass `pip-licenses --summary` which then provides output like so:

```bash
$ python ./piplicenses.py --summary
 Count  License
 1      ASL
 2      Apache 2.0
 8      BSD
 2      BSD License
 1      BSD or Apache License, Version 2.0
 1      BSD-like
 2      Expat license
 1      GPL
 2      LGPL
 14     MIT
 1      MIT License
 3      MIT license
 1      MPL-2.0
 1      PSF
 2      UNKNOWN
 1      http://opensource.org/licenses/MIT
 1      public domain, Python, 2-Clause BSD, GPL 3 (see COPYING.txt)
```

All output `--format-*` flags work with `--summary`, as do `--from-classifier`, `--with-system`, `--order` and `-i`. However, due to the nature of the summary only, `--with-authors` and `--with-urls` cannot take affect when `--summary` has been activated.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -344,7 +344,7 @@ def create_parser():
     parser.add_argument('--summary',
                         action='store_true',
                         default=False,
-                        help='show license use summary only')
+                        help='dump summary of each license')
 
     return parser
 
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -59,10 +59,22 @@ FIELD_NAMES = (
 )
 
 
+SUMMARY_FIELD_NAMES = (
+    'License',
+    'Count'
+)
+
+
 DEFAULT_OUTPUT_FIELDS = (
     'Name',
     'Version',
     'License',
+)
+
+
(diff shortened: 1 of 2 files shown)


raimon49:
@jnhmcknight Thank you to add a good implementation and perfect test.

I will merge this patch and release the next version to PyPI in the near future.

piplicenses.py
```diff
@@ -296,6 +341,10 @@ def create_parser():
                         action='store_true',
                         default=False,
                         help='dump as json')
+    parser.add_argument('--summary',
+                        action='store_true',
+                        default=False,
+                        help='show license use summary only')
```
raimon49: Can you change it like this?: `help='dump summary of each license'`

jnhmcknight: Sure thing.

raimon49:
@jnhmcknight Thanks!!

PR #16, opened 2018-12-22
Add --with-summary option

dtatarkin:
This option was very helpful to me. Hope this helps someone else.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -55,6 +55,7 @@ FIELD_NAMES = (
     'Version',
     'License',
     'Author',
+    'Summary',
     'URL',
 )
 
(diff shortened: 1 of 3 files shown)


piplicenses.py
```diff
@@ -310,6 +316,10 @@ def create_parser():
                         action='store_true',
                         default=False,
                         help='dump with package urls')
+    parser.add_argument('-d', '--with-summary',
```
raimon49: @dtatarkin Thanks for your patch. I want to merge because it looks good.

How about replacing the command line option with `--with-description` instead of `--with-summary`? It seems to be consistent with the short option `-d`.

dtatarkin: Yes, you are right `--with-description` is more intuitive, I will rename it

raimon49: great!! Thank you to add the perfect change and test. I will soon register the next release with PyPI.

Issue #18, opened 2019-02-08
Feature request: add option for limiting items from provided list


Liam-Deacon:
It would be great that there was an option to only list dependencies provided by a `requirements.txt` file or from the list of dependencies listed from the output of `pydeps --externals`.

PR #19, opened 2019-02-08
Added --packages option

Liam-Deacon:
Implemented as solution of #18


raimon49:
@Lightslayer Thank you for your feature request and implementation.

I think that this option is a good proposal.

In order to merge we have to solve 2 tasks.

1) I am against arguments receiving external files.This is because we need to implement a merge algorithm when multiple `.txt` or `.json` are given. It seems not to be taken into account in the current Pull Request. If this option is only accept the package name in the argument, it is acceptable to me.
2) pip-licenses always measures code coverage for quality. If you implement the new feature, please also write unit test in [test_piplicenses.py](https://github.com/raimon49/pip-licenses/blob/master/test_piplicenses.py). Tests can be run with `make test` command by constructing dependent environment with [dev-requirements.txt](https://github.com/raimon49/pip-licenses/blob/master/dev-requirements.txt)

Lightslayer:
Thanks @raimon49 for your feedback - and apologies it has taken me about a year to return to this!

I've now added some tests to cover the new functionality. On the point of loading multiple .txt or .json files, I am simply using a `set`  of all of the packages to get the list of merged package names, which is simple and should be okay (version pinning elements of the *requirements*.txt files are ignored when loaded thanks to some regex magic). 

Hope this is now suitable for bringing into the main code repo - please just let me know if you'd like me to make any more changes or add anything further.

README.md
```diff
@@ -189,6 +189,38 @@ Package names of arguments can be separated by spaces.
...
+### Option: packages
+
+When executed with the `--packages` option, only the packages specified by the argument are placed into the list output.
+
+```bash
+(venv) $ pip-licenses --packages twine numpy
+ Name  Version  License
+ numpy 1.16.1   BSD
+ twine 1.12.1   MIT
+```
```
raimon49, line 201: The implementation does not work as of this document.
```bash
# error with this option
$ pip-licenses --packages twine numpy
usage: piplicenses.py [-h] [-v] [--from SOURCE] [-s] [-a] [-u] [-d] [-l] [--no-license-path] [-i PKG] [-p INCLUDES] [-o COL]
                      [-f STYLE] [--summary] [--output-file OUTPUT_FILE]
piplicenses.py: error: unrecognized arguments: numpy
```

Since `action='append'` is used, it must be specified multiple times. Please see [docs for argpase](https://docs.python.org/3.8/library/argparse.html).

```bash
# work with this option
$ pip-licenses --packages twine --packages numpy
```

The behavior of the existing option `--ignore-packages` is also a breaking change.

Is it difficult to achieve your function with a combination of `action='store'` and `nargs='+'`?

test_piplicenses.py
```diff
@@ -229,17 +242,71 @@ def test_with_license_file_warning(self):
         self.assertIn('best paired with --format=json', warn_string)
 
     def test_ignore_packages(self):
-        if 'PTable' in SYSTEM_PACKAGES:
-            ignore_pkg_name = 'PTable'
-        else:
-            ignore_pkg_name = 'prettytable'
+        ignore_pkg_name = 'pytest'
```
raimon49, line 245: Why was this change needed?

raimon49:
@Liam-Deacon I adopted a simpler solution, #99, and shipped pip-licenses 3.4.0.

You can use pipe and shell scripts to work with the content output by pydep that fits your idea. Probably this way is more "UNIX-like".

Thanks for your suggestions and attempts.

PR #20, opened 2019-02-13
Add option to pull license file path and contents

cpeel:
My organization needs the ability to pull the license file path and contents, similar to how npm [license-checker](https://www.npmjs.com/package/license-checker) does when using `--customPath licenseText`. This adds that functionality with the `--with-license-file` argument.

During testing I also found a missing development dependency, hence the updates to `dev-requirements.in` (and the re-computed `dev-requirements.txt`) in a separate commit.


cpeel:
After more testing, this isn't full-baked yet. Will close and resubmit when it is.

PR #21, opened 2019-02-13
Minor test cleanup

cpeel:
`assertEquals()` and `assertNotEquals()` are deprecated in favor of their singular cousins `assertEqual()` and `assertNotEqual()`.

More importantly, remove a test function that is overwriting the one before it (because they have the same name) and yet contains duplicate functionality of the one after it.


raimon49:
@cpeel Thank you for the great contribution!!

PR #27, opened 2019-04-06
Minor fixes to Docker stuff

jameslamb:
@raimon49 thanks for this AWESOME project!

 I use it in CI for a few projects where I want to be really careful about unwanted licenses like GPL sneaking into my dependency tree.

Because I use `pip-licenses` so much, I wanted to come into the project today and spend some time learning how the internals work.

I really like the Docker setup you have, and hope you'll consider this PR that makes two small changes to it.

* Docker supports the use of `ARG` for storing (basically) environment variables that are only used at build time. I recommend centralizing the path `/opt/piplicenses` (which is manually copied in multiple places) in an `ARG` to improve the readability of your `Dockerfile` and reduce the risk of copy-paste errors if contributors add to it in the future
* While running through the Docker example in the README, I found that the outputs have changed slightly thanks to newer versions of `Werkzeug` being published

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,14 +1,17 @@
 FROM python:3.7.2-alpine
 LABEL maintainer="raimon <raimon49@hotmail.com>"
 
-WORKDIR /opt/piplicenses
+ARG APPDIR=/opt/piplicenses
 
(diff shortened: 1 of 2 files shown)


raimon49:
@jameslamb Thank you for the good suggestions!

PR #33, opened 2019-05-10
Quoted CSV output

vnagendra:
Added quoted CSV formatted output. Useful when you want to copy/paste the output to an excel sheet. Tests are also added.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -246,6 +246,24 @@ class JsonPrettyTable(PrettyTable):
         return json.dumps(lines, indent=2, sort_keys=True)
 
 
+class CSVPrettyTable(JsonPrettyTable):
+    """PrettyTable-like class exporting to CSV"""
+
+    def get_string(self, **kwargs):
+
+        options = self._get_options(kwargs)
+        rows = self._get_rows(options)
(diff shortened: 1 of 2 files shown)


raimon49:
@vnagendra Thanks. CSV format is great idea. I will merge this patch.

Some work is required before the next release version.

* Solve unit test `UnicodeDecodeError`
* Meta-escaping double quotes according to the [CSV specification](https://tools.ietf.org/html/rfc4180)
  * The meaning of this is the following replace: `"` -> `""`
* Add documents for CLI users

I will be in charge of these works. Do you mind?

vnagendra:
> I will be in charge of these works. Do you mind?

No issues from my side. I am also able to help out later today as well if you'd like. Apologies for sending you a PR that fails unit tests 😞. That's just wrong. I did run a local test on 2.7 and made sure it passed and avoided `u` .. but clearly not enough. 

If you want to have me take another shot, please let me know - otherwise, I am completely fine with whatever changes you choose to make.

raimon49:
I have released version 1.15.0.
https://pypi.org/project/pip-licenses/1.15.0/

We can output in CSV format! Many thanks 😄

PR #43, opened 2019-11-13
feat: Add JSONLicenseFinder

AlexanderVanEck:
Hi 👋 

I took the liberty of adding a --json-license-finder option to pip-licenses. This makes pip-licenses a drop-in replacement for LicenseFinder (https://github.com/pivotal/LicenseFinder) which is used by LicenseManagement in Gitlab's AutoDevops.

The issue with LicenseFinder is that it can only scan dependencies up until python 3.5 and not 3.6. pip-licenses is way better in that regard! However if we want to use the bindings inside Gitlab CI we need the output of pip-licenses to be the same as LicenseFinder.

From reading the source it wasn't immediately clear if pip-licenses supports overriding the json format that is outputted, if it does please point me in the right direction!

To me this seems like a clean solution. Suggestions on how to improve this PR are very welcome, this is my first PR into this project!

Alexander

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Dump the software license list of Python packages installed with pip.
         * [Confluence](#confluence)
         * [HTML](#html)
         * [JSON](#json)
+        * [JSON LicenseFinder](#json-licensefinder)
         * [CSV](#csv)
         * [Deprecated options](#deprecated-options)
     * [Option: summary](#option-summary)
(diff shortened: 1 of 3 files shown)


raimon49:
@AlexanderVanEck Thank you for your contribution.

I've also used LicenseFinder. It is a cool idea to output compatible JSON. I will work positively on merging this patch.

One question, pip-licenses plans to drop support for Python 2.7 in 2020. Do you does not matter it?

AlexanderVanEck:
@raimon49 Thank you :) Let me know if I can help in any way!

This feature has no problems with dropping 2.7 support to my knowledge, it is essentially the same as the JSON output (with a few key name changes).

raimon49:
@AlexanderVanEck Sure.

README.md
```diff
@@ -292,6 +292,27 @@ When executed with the `--format=json` option, you can output list in JSON forma
...
+  },
+  {
+    "licenses": ["MIT"],
+    "name": "pytz",
+    "version": "2017.3"
+  }
+]
+
+```
+
```
raimon49, line 316: Great documentation 👍 

This Markdown file has a table of contents at the top. Can you add a link to the table of contents for this option?

alexandervaneck: Added :)

raimon49:
Thanks, I will merge with master branch and release the next version shortly.

raimon49:
I have released version 1.16.0 to PyPI.
https://pypi.org/project/pip-licenses/1.16.0/

We can now use `--format=jlf` option. 😉

PR #52, opened 2019-12-17
Allow using prettytable

jayvdb:
PTable and prettytable conflict, both using package name 'prettytable'
with one as prettytable.py and the other as prettytable/.
pip provides no protection against two packages conflicting like
these.

PTable hasnt maintained backwards compatability, and hasnt been
doing releases for a while.

This patch allows either to be present.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -42,8 +42,22 @@ try:
 except ImportError:
     from pip import get_installed_distributions
 from prettytable import PrettyTable
-from prettytable.prettytable import (FRAME as RULE_FRAME, ALL as RULE_ALL,
-                                     HEADER as RULE_HEADER, NONE as RULE_NONE)
+try:
+    from prettytable.prettytable import (
+        ALL as RULE_ALL,
+        FRAME as RULE_FRAME,
+        HEADER as RULE_HEADER,
+        NONE as RULE_NONE,
+    )
+    PTABLE = True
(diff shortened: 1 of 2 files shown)


jayvdb:
See https://github.com/jazzband/prettytable/issues/23 for more info.

jayvdb:
c.f. https://build.opensuse.org/package/show/home:jayvdb:py-new/python-pip-licenses , where I use a little sed to change `setup.py` to use PrettyTable instead, and all tests pass.  There is no way that PTable will be acceptable to linux distros, especially given that it is also unmaintained.

If there are no objections to this patch, pls advise if I should use `nocover` or (much harder..) create a test scenario to cover these new code paths.

raimon49:
@jayvdb Thanks for the patch and information.

I understand that you supported compatibility.

I plan to merge this patch with #51  in the next release.

jayvdb:
To make codecov happy, should use nocover or (much harder..) create a test scenario to cover these new code paths?

raimon49:
I have confirmed that it works fine in a pip environment where the prettytable package is installed as an alternative to PTable.

Thank you for great work!

raimon49:
I released as version 1.18.0
https://pypi.org/project/pip-licenses/1.18.0/

jayvdb:
Thanks.  Submitted to openSUSE main Python package collection https://build.opensuse.org/request/show/758780

PR #57, opened 2020-01-04
Use collections.Counter for counting

adamchainz:
[Counter](https://docs.python.org/3/library/collections.html#collections.Counter) implements the exact pattern used here.


raimon49:
Thanks for the better way.

PR #61, opened 2020-01-25
Add --format=plain-vertical

### LINKED ISSUE #49: Add --format=plain-vertical format
We used `pip-licenses --with-license-file --format=plain` to try to generate something like Angular CLI's https://angular.io/3rdpartylicenses.txt format.

The issue with `pip-licenses`'s plain format is that it still tries to line things up horizontally, which can make it difficult to read in a browser. Also we noticed that some license files with special characters broke `pip-licenses` attempt to properly align the licenses.

Could a new format by added called `--format=plain-vertical` that puts the information in a vertical format similar to Angular CLI's generated output? This would be very useful.

johnthagen:
Closes #49 

This matches the license style of Angular CLI (which calls [`license-webpack-plugin`](https://github.com/xz64/license-webpack-plugin) internally) and is more pleasing to view on horizontally-constrained environments (like some web pages) when license files are included.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -414,8 +414,21 @@ def get_sortby(args):
     return 'Name'
 
 
+def create_plain_vertical_output(args, output_fields):
+    output = ''
+    for pkg in get_packages(args):
+        for field in output_fields:
+            if field.lower() in pkg:
+                output += '{}\n'.format(pkg[field.lower()])
+        output += '\n'
+    return output
+
+
 def create_output_string(args):
     output_fields = get_output_fields(args)
+
+    if args.format == 'plain-vertical':
+        return create_plain_vertical_output(args, output_fields)
 
     if args.summary:
         table = create_summary_table(args)
@@ -566,8 +579,9 @@ def create_parser():
                         action='store', type=str,
                         default='plain', metavar='STYLE',
                         help=('dump as set format style\n'
-                              '"plain", "markdown", "rst", "confluence",\n'
-                              '"html", "json", "json-license-finder",  "csv"\n'
+                              '"plain", "plain-vertical" "markdown", "rst", \n'
+                              '"confluence", "html", "json", \n'
+                              '"json-license-finder",  "csv"\n'
                               'default: --format=plain'))
     parser.add_argument('--summary',
                         action='store_true',
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -320,6 +320,26 @@ class CSVPrettyTable(PrettyTable):
         return '\n'.join(lines)
 
 
+class PlainVerticalTable(PrettyTable):
+    """PrettyTable for outputting to a simple non-column based style.
+
+    When used with --with-license-file, this style is similar to the default
+    style generated from Angular CLI's --extractLicenses flag.
+    """
+
+    def get_string(self, **kwargs):
+        options = self._get_options(kwargs)
+        rows = self._get_rows(options)
+
+        output = ''
+        for row in rows:
+            for v in row:
+                output += '{}\n'.format(v)
+            output += '\n'
+
+        return output
+
+
 def factory_styled_table_with_args(args, output_fields=DEFAULT_OUTPUT_FIELDS):
     table = PrettyTable()
     table.field_names = output_fields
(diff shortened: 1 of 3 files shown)


piplicenses.py
```diff
@@ -414,9 +414,22 @@ def get_sortby(args):
     return 'Name'
 
 
+def create_plain_vertical_output(args, output_fields):
```
raimon49: It's better to define a class that extends `PrettyTable`  than to define a new function. For example, `JsonPrettyTable`, `CSVPrettyTable`.

test_piplicenses.py
```diff
@@ -285,6 +285,11 @@ def test_format_plain(self):
         self.assertEqual('+', table.junction_char)
         self.assertEqual(RULE_FRAME, table.hrules)
 
+    def test_format_plain_vertical(self):
+        format_plain_args = ['--format=plain-vertical']
+        args = self.parser.parse_args(format_plain_args)
+        create_output_string(args)
```
raimon49: Use assert methods in unit tests to check expected values.

raimon49:
When run in my environment, the output is very simple.

```bash
PTable
0.9.2
BSD (3 clause)
```

The npm community has [license-checker](https://www.npmjs.com/package/license-checker) and [licensify](https://www.npmjs.com/package/licensify) implementations. In my opinion I find it more useful to be like these implementations. What do you think?

For example, the following output would be useful:

```bash
PTable:
  version: 0.9.2
  license: BSD (3 clause)
```

johnthagen:
@raimon49 

> The npm community has license-checker and licensify implementations. In my opinion I find it more useful to be like these implementations. What do you think?

I was trying to match the output style of Angular (see example here: https://angular.io/3rdpartylicenses.txt). Arguably, I'm using this with `--with-license-file --no-license-path` to match Angular closer.

I was also trying to keep the implementation as simple as possible, but I don't have a super strong opinion on how it actually looks, just that it's all vertically output rather than in columns (so it reads better on narrow windows).

What do you think? Since the license files themselves flow across multiple lines I wonder how that indented format would look. Would we need to indent every line of the license file?

piplicenses.py
johnthagen: I actually did play around with `PrettyTable` for some time trying to get it to match the [Angular output](https://angular.io/3rdpartylicenses.txt). The problem I kept running into is that fundamentally `PrettyTable` is trying to create a horizontal table, but in this mode we don't want any of that (no columns, padding, separators, etc).

I found myself basically trying to hack `PrettyTable` to undo the "table" part of it and it was looking really ugly. I realized that for this really simple vertical format that just manually looping was easier to read and reason about.

Do you know how to get `PrettyTable` to do this?

raimon49: For example, can't you define it like this and get the behavior you want?

```python
class AngularStyledTable(PrettyTable):

    def get_string(self, **kwargs):
        options = self._get_options(kwargs)
        rows = self._get_rows(options)

        output = ''
        for row in rows:
            for v in row:
                output += '{}\n'.format(v)

            output += '\n'

        return output
```

In my opinion, inheriting `PrettyTable` like other contributors works better with options like `--sort` and `--with-urls` and more.

raimon49:
@johnthagen All right. No problem with Angular-like styles.

In addition, I require that this new option be specified in the `README.md` document.

johnthagen:
@raimon49 Great review, let me know what you think of the latest changes.

Please squash the commits if/when you merge. :)

raimon49:
@johnthagen Thanks for the good change 👍 

Please add it to the Table Of Contents in `README.md`.

> Please squash the commits if/when you merge. :)

Sure.

johnthagen:
> Please add it to the Table Of Contents in README.md.

👍 Done.

raimon49:
Thank you to all of the work.

We will be able to use this format in the next release version of pip-licenses.

johnthagen:
> We will be able to use this format in the next release version of pip-licenses.

Awesome! Do you have any plans for when that might be?

raimon49:
I will be uploading to PyPI this week.So you can use it on 2020-02-01.

PR #83, opened 2020-11-24
Added fail-on and allow-only arguments

dkarp0:
As used by https://www.npmjs.com/package/license-checker, these
arguments are useful when using pip-licenses in continuous integration
to prevent risky licenses.

fail-on argument can be used to as a disallow list to block select
unwanted licenses
allow-only argument can be used with an allow list of licenses that have
been pre-approved to force a review when a package with a new license is
added


piplicenses.py
```diff
@@ -220,6 +229,26 @@ def get_pkg_info(pkg):
             continue
 
         pkg_info = get_pkg_info(pkg)
+
+        license_name = pkg_info['license']
+        if fail_on_licenses and license_name in fail_on_licenses:
```
raimon49, line 234: In packages that adopt a dual license, the `license_name` is expressed as a comma-separated list. A comparison with an exact match may be made through.

e.g.) 
```sh
# Packages distributed under dual licenses
$ pip show -v keyring | grep "License :: OSI Approved"
  License :: OSI Approved :: Python Software Foundation License
  License :: OSI Approved :: MIT License

# That's comma-separated output 
$ pip-licenses | grep keyring 
 keyring             21.4.0     Python Software Foundation License, MIT License

# Just specify "Python Software Foundation License" and it went through
$ pip-licenses --fail-on="Python Software Foundation License"
$ echo $?
0
```

Is this as expected?

If it is as expected, it would be nice to have it described in more detail in the `README` documentation.

raimon49 (changes requested):
@dkarp0 Thanks for the patch. This is a great contribution.

I wrote one comment. How to handle dual licenses is a vexing issue for this tool.

I think it's better to merge the code as is than to make it more complex. I would like to enhance the `README` documentation.

piplicenses.py
dkarp0: Indeed, dual licenses is an issue we have using https://www.npmjs.com/package/license-checker also where `MIT or GPL` is distinct from `MIT`.

This is less of an issue for `allow-only` as you can add `MIT or GPL` and `MIT` when you see the check is failing. It means you are forced to approve the new license combination. This is the form I need most.

However for `fail-on` this may be more of an issue. For example you may be trying to specifically prevent `GPL` and as in your example, this would not match if there was another license also attached. I agree that a stern warning in the README would be a good start and then a possible improvement could come later, I'll do that now ;)

raimon49 (approved):
Excellent!!

I will merge this patch and release the next version 3.1.0.

dkarp0:
Thanks for the quick merge @raimon49 - I didn't expect I'd be able to switch back over from my fork already today!

Issue #72, opened 2020-07-04
Adopt option for SPDX-License-Identifiers output


alandtse:
Great project! I handle open source compliance for a large multinational corporation and just found this while coding in my spare time. I'll probably start pointing my developers to this project for python license identification from their dependencies.

It'd be nice if there was an option to use SPDX-License-Identifiers for the output for licenses. This is the identifier based on this [list](https://spdx.org/licenses/) and is intended to normalize identification of licenses.  

These identifiers are actually being used in multiple projects now to identify the license (e.g., Linux [kernel](https://www.kernel.org/doc/html/v4.18/process/license-rules.html)). 

This may complicate how this tool works since it looks like you're just passing the meta data through, but I think more python projects may adopt the `SPDX-License-Identifier` format.

raimon49:
@alandtse Hey, thanks for the great suggestions.

Support for SPDX IDs is certainly beneficial.

I don't have time to work on this suggestion right now. But it may support it in the future.

If the behavior spec is written in this issue, it will be easier to work on implementation.

Alternatively, you are welcome to provide patches.

stefan6419846:
In my opinion, this conversion should probably be implemented in a dedicated package instead of the `pip-licenses` package. From my experience, there are quite some edge cases or strange license field usages, which have to be considered - let alone the list of licenses itself which just introduces overhead for everyone who does not need this feature. (Example: I regularly have a look a the license fields of about 200 PyPI packages extracted by `pip-licenses` and observed quite some oddities, including links to Wikipedia etc.)

Providing a Python-based interface as requested in #81 should allow everyone to plug the desired license format conversion tooling into the tool chain (although this can already be done now when `pip-licenses` is being called correctly).

raimon49:
> Providing a Python-based interface as requested in #81 should allow everyone to plug the desired license format conversion tooling into the tool chain (although this can already be done now when `pip-licenses` is being called correctly).

I agree with you. When a user wants to achieve a slightly nifty conversion or output, it is much easier than sending a patch to `pip-licenses`.

If we provide a Python code-based interface, we must work on preparing and documenting type information.

anthonyharrison:
I think this is an essential feature.

[...]

PR #86, opened 2020-12-20
Added '--from=all' option

cdce8p:
At the moment `pip-licenses` can only output **either** the metadata **or** the classifier license information.
This PR adds the `--from=all` option that displays both license information at the same time .

**Background**
[PEP 639](https://www.python.org/dev/peps/pep-0639/) specified how Python packages are supposed to define license information going forward. The preferred way is to use the `license` metadata field with SPDX license expression [syntax](https://www.python.org/dev/peps/pep-0639/#license-expression-syntax). Although not officially deprecated, the trove classifiers will slowly be phased out and won't receive any more updates.
Since the PEP is only from Aug-2019, most packages still use the trove class and/or have outdated/invalid metadata license information. With the proposed option, it will be easier to compare the used licenses.

**Going forward**
- `--from=mixed` still prefers the classifier over metadata information. Changing this would be a big breaking change and probably confuse many users. Therefore I'm not sure if it will be worth it. At least not for now.
- Update the metadata `license` information for `pip-licenses` to just `MIT` to conform with the SPDX identifier.
- Try parsing the metadata license field and extracting SPDX identifiers.

**Ideas**
- Add option to auto-convert classifier to SPDX identifier (when possible, eg. only in simple cases)
- Parse SPDX license expression to better evaluate `fail` / `allow`
- Add license pinning, would be useful for CI integrations to require manual approval if license for specific package changes. This would probably work best with a config file.

--

## Next steps for this PR
- [x] Finalize implementation details
- [x] Add test case(s)

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -92,7 +92,6 @@ SUMMARY_FIELD_NAMES = (
 DEFAULT_OUTPUT_FIELDS = (
     'Name',
     'Version',
-    'License',
 )
 
 
@@ -107,12 +106,15 @@ METADATA_KEYS = (
     'author',
     'license',
     'summary',
+    'license_classifier',
 )
 
 # Mapping of FIELD_NAMES to METADATA_KEYS where they differ by more than case
 FIELDS_TO_METADATA_KEYS = {
     'URL': 'home-page',
     'Description': 'summary',
+    'License-Metadata': 'license',
+    'License-Classifier': 'license_classifier',
 }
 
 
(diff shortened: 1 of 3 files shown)


raimon49:
@cdce8p Thanks for the info on PEP 639. I will review and feedback on your branch at a later date.

> - Update the metadata license information for pip-licenses to just MIT to conform with the SPDX identifier.

Yes, this will be a task I will tackle when I release the next version 😉

raimon49:
This discussion is also related to issue #72 .

piplicenses.py
```diff
@@ -432,20 +445,16 @@ def find_license_from_classifier(message):
             if license != 'OSI Approved':
                 licenses.append(license)
 
-    if len(licenses) > 0:
-        license_from_classifier = ', '.join(licenses)
-
-    return license_from_classifier
+    return licenses
```
raimon49, line 448: Just my little opinion, the return value of the function `find_license_from_classifier` was changed from `str` to `list`, which made other implementation and test codes more complicated.

But it's not important enough to oppose the merge.

raimon49:
@cdce8p I've checked your patch. I'm glad you worked with attention to compatibility.

I am positive that I will merge this patch. I hope to move on to the next step.

piplicenses.py
cdce8p: You're right that it made the commit a bit more complicated. However I believe this is the better solution here. Changing the responsibility of this function to only extract the license classifier might make the work down the line a bit easier.

cdce8p:
Thanks for taking the time to review my changes. I went ahead and added a test case for the change.
The PR should be good on my part now. Let me know if there is anything else I should do.

raimon49 (approved):
Thank you for your contribution.

This feature will be in the next release, 3.2.0.

raimon49:
Version 3.2.0 has just been released!
https://pypi.org/project/pip-licenses/3.2.0/

Issue #82, opened 2020-11-24
Switch to PrettyTable


jayvdb:
PrettyTable is now maintained by jazzband and a new release is out.

raimon49:
Hmm, that's a breaking change to be considered in the next major release.

johnthagen:
I came here to suggest the same thing. Since `PTable` hasn't seen a release since 2015, it's basically abandoned. It is nice to see `prettytable` alive again: https://github.com/jazzband/prettytable

raimon49:
I was just following the implementation of the `prettytable` package for another issue and noticed one thing.

In the current version of prettytable 2.5.0, the number of dependent packages has increased.

https://github.com/jazzband/prettytable/blob/2.5.0/setup.cfg
```
[options]
install_requires =
    wcwidth
    importlib-metadata;python_version < '3.8'
```

So, after switching from `PTable` to `prettytable`, we need to decide whether we should add more system packages?

ref) #119 #117 #89

johnthagen:
> So, after switching from PTable to prettytable, we need to decide whether we should add more system packages?

This is a good point... Ideally there wouldn't need to be extra system packages, but I do think that anything that `pip-licenses` depends on should be excluded by default since it will be extremely likely that the end user isn't actually depending on them.

Perhaps the additional dependencies are a motivation to stay with `PTable` if it's not causing issues? How would you rate the current maintenance burden of `PTable` not being maintained?

Then there is the extra question of what happens when `prettytable` decides to add extra dependencies, some of which could be common?

raimon49:
> Perhaps the additional dependencies are a motivation to stay with `PTable` if it's not causing issues? How would you rate the current maintenance burden of `PTable` not being maintained?

Yes, in the case where `pip-licenses` outputs a license list, I would rate `PTable` as stable enough. However, we may have an unknown bug, or it may not work in future Python versions.

> Then there is the extra question of what happens when `prettytable` decides to add extra dependencies, some of which could be common?

I read a bit about how the dependency was added to `prettytable`. It was to better handle Unicode, such as CJK. As I am also a Japanese speaker, I understand this addition. However, many open source licenses are written in English, which is a feature that `pip-licenses` do not need. `prettytable` is being developed on a community basis, and more features will be added in the future.

I have one idea and solution. That's vendoring original `PTable` code in `pip-licenses` and distributes it.

```diff
-from prettytable import PrettyTable
+from ._vendor.prettytable import PrettyTable

[...]

Issue #81, opened 2020-11-06
Calling from within Python


fladd:
The documentation currently does not explain how to call pip-licenses from within Python, only how to call it from a command line.

raimon49:
Yes, pip-licenses is developed as a CLI tool.

Does your opinion mean that you want to import pip-licenses as a module and use some of the features in your Python code?

fladd:
Yes, that would be great. Currently I get around it by calling it with subprocess and then parsing the json output, but this seems a bit unnecessary, given that pip-licenses is Python code itself.

raimon49:
I didn't imagine anyone would want to use the scan results in their Python code.

I would consider APIs and documentation that users can handle with structured objects.

Thank you for your valuable opinion, @fladd .

sweigert:
> Yes, that would be great. Currently I get around it by calling it with subprocess and then parsing the json output, but this seems a bit unnecessary, given that pip-licenses is Python code itself.

This is what I am doing to achieve this:
```python
import piplicenses
parser = piplicenses.create_parser()
args = parser.parse_args()
packages = piplicenses.get_packages(args)
for p in packages:
   print(p)
```

diazona:
This would also be useful to work around the lack of "proper" support for multiply-licensed packages - that is, the fact that multiple licenses are shown in a single string separated by delimiters, rather than as some structured representation (like a JSON list, for the JSON output format). If there were a Python API, perhaps it could expose multiple licenses as a Python list.

I don't know how practical this actually is, but it would make a significant difference to my use case.

bartfeenstra:
I'd like to express my support for this feature request. I'm currently using pip-licenses in a unit test to ensure all dependencies' licenses are compatible. A pip-licenses Python API would allow this code to be rewritten a lot shorter and cleaner.

pwuertz:
A pip-licenses Python API would be a great fit for applications bundled with e.g. PyInstaller.

Using `piplicenses` within a PyInstaller-specfile, you could fully integrate the licence collection process right into your application bundling process.

stefan6419846:
I previously used `packages = list(piplicenses.get_packages(from_source=piplicenses.FromArg.MIXED))` to get the packages from within Python without relying on a subprocess.

[...]

PR #87, opened 2021-01-18
Improve argparser

cdce8p:
Some improvements to the argument parser. This PR greatly improves the readability of the help command and introduces `Enums` for arguments.

**Current help text**
Screen Shot 2021-01-19 at 00 54 59

**Changes**
- `\n` in help strings are now rendered correctly
- Added argument groups and rearranged argument positions
- Updated readme to match changes
- Updated `max_help_position` to avoid unnecessary line breaks
- Added choices as additional help for `from`, `order` and `format`
- Added enums for `from`, `order` and `format`

Screen Shot 2021-01-19 at 00 39 34

Screen Shot 2021-01-19 at 00 57 35

--

I've split these changes into multiple commits to make reviewing them a bit easier.

On a related note, what is your opinion to `squash merge` for PRs? Although it might seem counterintuitive at first, I think it makes the git log much easier to read and the Github integration is better. The squash commit message includes the PR number which will be automatically be linked (for `git blame`).

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -234,7 +234,7 @@ def get_packages(args: "CustomNamespace"):
         pkg_info = get_pkg_info(pkg)
 
         license_name = select_license_by_source(
-            getattr(args, 'from'),
+            args.from_,
             pkg_info['license_classifier'],
             pkg_info['license'])
 
@@ -262,14 +262,13 @@ def get_packages(args: "CustomNamespace"):
 def create_licenses_table(
         args: "CustomNamespace", output_fields=DEFAULT_OUTPUT_FIELDS):
     table = factory_styled_table_with_args(args, output_fields)
-    from_source = getattr(args, 'from')
 
     for pkg in get_packages(args):
         row = []
         for field in output_fields:
             if field == 'License':
                 license_str = select_license_by_source(
-                    from_source, pkg['license_classifier'], pkg['license'])
+                    args.from_, pkg['license_classifier'], pkg['license'])
                 row.append(license_str)
             elif field == 'License-Classifier':
                 row.append(', '.join(pkg['license_classifier'])
@@ -412,26 +411,26 @@ def factory_styled_table_with_args(
     table = PrettyTable()
     table.field_names = output_fields
     table.align = 'l'
-    table.border = args.format in (FormatArg.MARKDOWN, FormatArg.RST,
-                                   FormatArg.CONFLUENCE, FormatArg.JSON)
+    table.border = args.format_ in (FormatArg.MARKDOWN, FormatArg.RST,
+                                    FormatArg.CONFLUENCE, FormatArg.JSON)
     table.header = True
 
-    if args.format == FormatArg.MARKDOWN:
+    if args.format_ == FormatArg.MARKDOWN:
         table.junction_char = '|'
         table.hrules = RULE_HEADER
-    elif args.format == FormatArg.RST:
+    elif args.format_ == FormatArg.RST:
         table.junction_char = '+'
         table.hrules = RULE_ALL
-    elif args.format == FormatArg.CONFLUENCE:
+    elif args.format_ == FormatArg.CONFLUENCE:
         table.junction_char = '|'
         table.hrules = RULE_NONE
-    elif args.format == FormatArg.JSON:
+    elif args.format_ == FormatArg.JSON:
         table = JsonPrettyTable(table.field_names)
-    elif args.format == FormatArg.JSON_LICENSE_FINDER:
+    elif args.format_ == FormatArg.JSON_LICENSE_FINDER:
         table = JsonLicenseFinderTable(table.field_names)
-    elif args.format == FormatArg.CSV:
+    elif args.format_ == FormatArg.CSV:
         table = CSVPrettyTable(table.field_names)
-    elif args.format == FormatArg.PLAIN_VERTICAL:
+    elif args.format_ == FormatArg.PLAIN_VERTICAL:
         table = PlainVerticalTable(table.field_names)
 
     return table
@@ -465,7 +464,7 @@ def get_output_fields(args: "CustomNamespace"):
 
     output_fields = list(DEFAULT_OUTPUT_FIELDS)
 
-    if getattr(args, 'from') == FromArg.ALL:
+    if args.from_ == FromArg.ALL:
         output_fields.append('License-Metadata')
         output_fields.append('License-Classifier')
     else:
@@ -519,7 +518,7 @@ def create_output_string(args: "CustomNamespace"):
 
     sortby = get_sortby(args)
 
-    if args.format == FormatArg.HTML:
+    if args.format_ == FormatArg.HTML:
         return table.get_html_string(fields=output_fields, sortby=sortby)
     else:
         return table.get_string(fields=output_fields, sortby=sortby)
@@ -529,7 +528,7 @@ def create_warn_string(args: "CustomNamespace"):
     warn_messages = []
     warn = partial(output_colored, '33')
 
-    if args.with_license_file and not args.format == FormatArg.JSON:
+    if args.with_license_file and not args.format_ == FormatArg.JSON:
         message = warn(('Due to the length of these fields, this option is '
                         'best paired with --format=json.'))
         warn_messages.append(message)
@@ -583,9 +582,9 @@ class CustomHelpFormatter(argparse.HelpFormatter):  # pragma: no cover
 
 
 class CustomNamespace(argparse.Namespace):
-    _from: "FromArg"
+    from_: "FromArg"
     order: "OrderArg"
-    format: "FormatArg"
+    format_: "FormatArg"
     summary: bool
     output_file: str
     ignore_packages: List[str]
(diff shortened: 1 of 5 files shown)


raimon49:
@cdce8p Wow! Thank you. It's so easy to read!

I didn't know that you can do this hack with `add_argument_group` and self defined `HelpFormatter`. It's so cool.

> what is your opinion to `squash merge` for PRs?

I like to commit to smaller units. That helps me in my review as well. Thanks.

I don't always squash merge approved PRs, but there is nothing wrong with squash merging this PR.

piplicenses.py
```diff
@@ -563,150 +601,190 @@ def _check_code_page(code_page):
...
-                        action='store', type=str,
-                        default=None,
-                        help='fail (exit with code 1) on the first occurrence '
-                             'of the licenses not in the semicolon-separated '
-                             'list')
+        description=__summary__,
+        formatter_class=CustomHelpFormatter)
+
+    common_options = parser.add_argument_group('Common options')
+    format_options = parser.add_argument_group('Format options')
```
raimon49, line 703: The argument_group title "Format options" looks a bit strange. Because option ` --format` is in "Common options".

But I don't have a good alternative either 😅 

By the way, "Verify options" is a great naming 👍

cdce8p: I agree, it's a bit strange. I can change the group names if you like, although I think it's good enough. The goal with `Format options` was to group all **additional** format arguments that one might not need that often.

cdce8p:
I just remembered that I missed implementing a few small checks. I'll add them later today.
Those checks should cover specifying `--no-license-path` and `--with-notice-file` without also adding `--with-license-file` **and** `--filter-code-page` without `--filter-strings`.

cdce8p:
Just pushed all remaining changes. In addition to extending args verification, I've also added some type hints for `args`. Those allow for much better autocompletions in IDEs. Lastly, I added a basic config for `isort` in `setup.cfg`. With that it's pretty easy to sort all imports and make them look nice.

After this PR is finished, I would like to propose moving from a single source file to a package structure. `piplicenses.py` is already quite large with almost 900 lines and it would only continue to grow in the future. The move could improve the readability. What do you think?

setup.cfg
```diff
@@ -49,3 +49,10 @@ test = pytest
 
 [tool:pytest]
 addopts = --pycodestyle -v --cov --cov-report term-missing
+
+[tool:isort]
+# https://github.com/timothycrosley/isort
+multi_line_output = 4
+line_length = 72
+known_first_party =
+    piplicenses
```
raimon49, line 58: @cdce8p Thank you for your great contribution. This is a very useful tool.

I have something to ask for future reference. Why do you want to overwrite the `line_length` setting from 79 (as default) to 72? I'd like to hear if there is any reason to recommend it.

raimon49:
> After this PR is finished, I would like to propose moving from a single source file to a package structure. piplicenses.py is already quite large with almost 900 lines and it would only continue to grow in the future. The move could improve the readability. What do you think?

Yes, I agree with your proposal.

We have a lot of unit test code. It will also help when changing to a package structure. The test code you added is one of them.

setup.cfg
cdce8p: To be honest, I don't really know. It's something I have picked up somewhere over time and it seems to work great. I guess an advantage would be that with only `72` characters not to many `from` imports are on one line, see `test_piplicenses.py` for example.

Regarding the default `79`: Personally I would recommend to change it. I tend to use `119` for my projects. That doesn't mean that every line should now be longer, but that in some **limited** cases it improves readability when you otherwise would be only a few characters over the limit. In fairness a line that approaches 119 would probably be best split up anyway, but it's our choice if it's really the best option, not of the style checker.

raimon49: > To be honest, I don't really know. It's something I have picked up somewhere over time and it seems to work great. I guess an advantage would be that with only `72` characters not to many `from` imports are on one line, see `test_piplicenses.py` for example.

Okay, I understand. Indeed, it seems to be working well.

> That doesn't mean that every line should now be longer, but that in some **limited** cases it improves readability when you otherwise would be only a few characters over the limit. 

In limited cases, you're right. I'm following the Python community standard for now.

raimon49 (approved):
I checked this P-R.

It will be imported and shipped in the next release version 3.3.0. The work is scheduled for the weekend around 2021-01-23.

cdce8p:
That's also where squash merge has the most benefits.

Just to clarify, how the whole process would look like:
1. **Squash merge** every new PR into `dev`
2. For new release
  2.1. Push new commit updating the version string
  2.2. Create new PR, `dev` against `master`, this will run all CI on last time
  2.3. If CI passes, **normal merge** PR into `master`
  2.4. Since `master` now contains an additional merge commit that `dev` doesn't have, merge `master` back into `dev`. A normal push is enough
3. Go back to step **1**

Some additional notes:
- If you set the default branch to `dev`, all new PRs will target it by default
- All changes, except version updates, should only be **squash merged** via a PR. This guarantees that all commits which show up in git blame reference a PR with additional information
- If all changes have a PR, it also makes sure that all checks have passes and you don't accidentally committed something which causes a bug.
- You could add the requirement that all checks must pass before a PR can be merged
- Optional add that every PR needs at least one approving review
- I believe it's also possible to set branch protections in Github, eg. for `master` don't allow push commits

--
Please let me know if you have any questions

raimon49:
To express my personal opinion:

I am against having a permanent `dev` branch. You are right, it is profitable for big projects to develop several new features in parallel. But I do not see this tool as a big project. It is better to keep the flow simple, as shown in the [GitHub flow](https://guides.github.com/introduction/flow/).

Next, let's talk about the release plan.

Improve argparser is a simple and beautiful advantage for many users.

On the other hand, in the process of embarking on the move to the package structure, it is possible that some of the most recent issues that have arrived can be resolved. (e. g. #82 #81 )

It seems to me that this effort is worth working on the base branch as you suggest. For example, create a base branch such as `dev-4.0.0` or `release-4.0.0`. However, as I mentioned earlier, I am against a permanent branch. So, when the goal of the package structure is achieved and shipped, this development branch will end its lifecycle.

I'm waiting for your opinion.

cdce8p:
That works for me. My intention behind the suggestion was simply to allow for a faster development pace. Implementing all changes would otherwise either require one huge PR, which would basically be unreviewable, or take months to complete. Having a separate branch thus allows for a faster turnaround once you approve a PR. I don't really have a name preference.

Regarding the issues you mentioned, it's definitely worth keeping them in mind. I also mentioned some other topics in my last PR #86 which I would like to work on. Targeting `4.0` would certainly make things easier, regarding breaking changes.
I plan on opening a new issue later to track all topics for an upcoming release. This would also be a good place to have general discussions if topics pop up.

cdce8p:
While I was working on #88, I noticed some small issues. The [last commit](https://github.com/raimon49/pip-licenses/pull/87/commits/b7949b5f5b739ecb1777da850f6756a6fac8af6a) address these.

raimon49:
@cdce8p Am I correct in understanding that your work on the `improve-argparser` branch is complete?

Then I will start the next task. Please let me know if there is any miscommunication.

1. Merge this P-R (with squash)
2. Version 3.3.0 will be shipped based on the master branch.
3. Create a remote branch `dev-4.0.0` .

cdce8p:
@raimon49 That is correct. This PR is complete.

cdce8p:
@raimon49 Something went wrong with the squash merge. Did you do it with Github?
To be honest, I would like to see it fixed. I could post some commands that should do it if you agree.

cdce8p:
The following should work:
1. Disable branch protection for `master` (if enabled)
2. Local git commands
```
git fetch origin
git checkout master
git rebase origin/master
git checkout -b master-old
git checkout master
git reset --hard HEAD~2
git push -f
```
3. Reopen #87 + from Github select and do squash merge (into master)
4. Local git commands
```
git fetch origin master
git rebase origin/master
git cherry-pick c2da089e63b47121cddf73dcc85695a29bb14570  # Bump version
git cherry-pick 3634bab33900f30e2863a5256f906c24289937cd  # isort
git cherry-pick 4ace27c31978f42f77b979f90549277f94986ac3  # make target 'format'
git cherry-pick -m 1 a5cc189f88f50095533d98d0ec8230229b9d2ca2   # merge PR #90
git commit --allow-empty
git push

git checkout dev-4.0.0
git reset --hard master
git push -f

git branch -D master-old
```
5. Reenable branch protection for `master`

raimon49:
@cdce8p I worked on it.

cdce8p:
Looks like it worked. Thanks!

cdce8p:
@raimon49 Just one last thing: Do you want to release this as `3.3.1`? That way we could guarantee this issue shouldn't pop up anywhere in the future.

raimon49:
I have released 3.3.0 to PyPI based on commit 4ace27c31978f42f77b979f90549277f94986ac3.

As far as I can see, there is no difference from the master branch that was forced to be pushed.
```bash
$ git diff origin/master 4ace27c31978f42f77b979f90549277f94986ac3
```

I have determined that users who have the package installed will not be affected. So I decide that there is no need to release 3.3.1.

PR #88, opened 2021-01-21
Move to package structure

cdce8p:
This PR is based on the discussion in #87
I've split up the code into multiple files. Additionally I refactored the tests to use pytest instead of unittest.

--
Until #87 is merged, this PR contains a commit with those changes. I will remove it once done.


raimon49:
I've prepared the following remote branch.
https://github.com/raimon49/pip-licenses/tree/dev-4.0.0

It would be good to consolidate the results of this P-R into a `dev-4.0.0` branch.

cdce8p:
I rebased the branch onto `dev-4.0.0`, so it's ready for review.

I will however leave it in `Draft` status until the issue with the squash commit is resolved: https://github.com/raimon49/pip-licenses/pull/87#issuecomment-765897656
That will only require a rebase, so the code itself won't change.

cdce8p:
I've rebased the branch. The PR is ready for review now

cdce8p:
@raimon49 Did you get a chance to take a look at this yet?

.github/workflows/python-package.yml
```diff
@@ -24,9 +24,10 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
+        pip install -e .
```
raimon49, line 31: Does this mean that the local pip-licenses package installation will be tested for success?

piplicenses/const.py
```diff
@@ -0,0 +1,128 @@
...
+
+
+# General information
+__pkgname__ = 'pip-licenses'
+__version__ = '3.3.0'
+__author__ = 'raimon'
+__license__ = 'MIT'
+__summary__ = ('Dump the software license list of '
+               'Python packages installed with pip.')
+__url__ = 'https://github.com/raimon49/pip-licenses'
```
raimon49, line 22: It's a good idea that the constants for the whole package are now in `const.py` 👍 

On the other hand, it seems that meta-information such as `__version__` is often declared in `__init__.py`.

What do you think?

raimon49:
@cdce8p Overall, I am willing to distribute pip-licenses under the MIT license.

So, I would like all `piplicenses/*.py` files to have a comment with the license clause.

.github/workflows/python-package.yml
cdce8p: I don't expect `pip install -e .` to fail. However this is necessary for relative imports inside the package.

piplicenses/const.py
cdce8p: I've seen it being declared in `const.py` as well. Personally I like it a bit more that way. It's easier to use use these in the remaining code. Eg. the argparser `-v` command.

cdce8p:
> @cdce8p Overall, I am willing to distribute pip-licenses under the MIT license.
> 
> So, I would like all `piplicenses/*.py` files to have a comment with the license clause.

I agree, the MIT license is a really good fit for this project. However I don't believe it's really necessary to add it to every file header. In almost all cases it's enough to have a [LICENSE](https://github.com/raimon49/pip-licenses/blob/master/LICENSE) file distributed with the package. Doing it that way, it's also way easier to update the copyright notice (eg. to update the year). The only cases I can think of including the license in every file use the `GPL License` and even then I personally would advise against it.

.github/workflows/python-package.yml
raimon49: I got it. Since "Install dependencies" has already been completed, the following command may be fine.

```bash
pip install -e . --no-deps
```

piplicenses/const.py
raimon49: Okay, let's just let it be declared in `const.py`.

tests/test_main.py
```diff
@@ -0,0 +1,41 @@
...
+
+@pytest.mark.parametrize('order_args,field_name', [
+    (['--order=name'], 'Name'),
+    (['--order=license'], 'License'),
+    (['--order=author', '--with-authors'], 'Author'),
+    (['--order=url', '--with-urls'], 'URL'),
+    (['--order=url'], 'Name'),
+    (['--summary', '--order=count'], 'Count'),
+    (['--summary', '--order=name'], 'License'),
+])
```
raimon49, line 19: The use of parametrizing test looks good. 👍

raimon49 (approved):
The split design of the package structure looks like a good job.

After this PR is merged, or just before v-4.0.0 ships, I may add a simple link to the license file URLs and a declaration of character encoding to all files.

.github/workflows/python-package.yml
cdce8p: I don't think it's necessary. pip automatically checks if the dependencies are installed, thus `pip install -e .` will only link the `piplicenses` package.

See the [action log](https://github.com/raimon49/pip-licenses/pull/88/checks?check_run_id=1800977022) -> `Install dependencies` -> L124.

cdce8p:
> After this PR is merged, or just before v-4.0.0 ships, I may add a simple link to the license file URLs and a declaration of character encoding to all files.

I don't think it's necessary, but I can understand your intension. I've added a short note to the top of every source file:
```py
# ---------------------------------------------------------------------------
# Licensed under the MIT License. See LICENSE file for license information.
# ---------------------------------------------------------------------------
```
What do you think?

--

**Character encodings**: I believe they aren't necessary. `utf-8` is the default in Python 3. See [stackoverflow](https://stackoverflow.com/questions/4872007/where-does-this-come-from-coding-utf-8/58652741#58652741). Vim doesn't need it either.

--

`#!/usr/bin/env python`: Also not necessary, since the python files aren't executed directly.

cdce8p:
Can you squash merge this PR (and all other subsequent ones for the dev branch) as well?

.github/workflows/python-package.yml
raimon49: My suggestion means that we can skip L123. But it is a trivial matter. We'll be fine as is.

raimon49:
@cdce8p 

> I've added a short note to the top of every source file:

Thanks for the work. It looks good.

> **Character encodings**: I believe they aren't necessary. `utf-8` is the default in Python 3. See [stackoverflow](https://stackoverflow.com/questions/4872007/where-does-this-come-from-coding-utf-8/58652741#58652741). Vim doesn't need it either.

Okay, I understand.

raimon49:
> Can you squash merge this PR (and all other subsequent ones for the dev branch) as well?

Yes, we will adopt that workflow for the merge to `dev-4.0.0`.

cdce8p:
> > Can you squash merge this PR (and all other subsequent ones for the dev branch) as well?
> 
> Yes, we will adopt that workflow for the merge to `dev-4.0.0`.

Thanks 👍🏻

PR #102, opened 2021-06-28
Verify multiple licenses individually

khatkar:
* Changed behavior of `allow-only` and `fail-on` flags. Now you do not
  need to specify exact combinations of multiple licenses.
* Updated Readme and tests

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -237,34 +237,33 @@ def get_packages(args: "CustomNamespace"):
 
         pkg_info = get_pkg_info(pkg)
 
-        license_name = select_license_by_source(
+        license_names = select_license_by_source(
             args.from_,
             pkg_info['license_classifier'],
             pkg_info['license'])
 
-        license_names = set(license_name.split(', '))
-
         if fail_on_licenses:
             failed_licenses = license_names.intersection(fail_on_licenses)
             if failed_licenses:
-                sys.stderr.write("fail-on license {} was found for package "
-                                 "{}:{}".format(
-                                    ', '.join(failed_licenses),
-                                    pkg_info['name'],
-                                    pkg_info['version'])
-                                 )
+                sys.stderr.write(
+                    "fail-on license {} was found for package "
+                    "{}:{}".format(
+                        '; '.join(failed_licenses),
+                        pkg_info['name'],
+                        pkg_info['version'])
+                )
                 sys.exit(1)
 
         if allow_only_licenses:
-            missing_licenses = license_names.difference(allow_only_licenses)
-            if missing_licenses:
+            uncommon_licenses = license_names.difference(allow_only_licenses)
+            if len(uncommon_licenses) == len(license_names):
                 sys.stderr.write(
                     "license {} not in allow-only licenses was found"
                     " for package {}:{}".format(
-                       ', '.join(missing_licenses),
-                       pkg_info['name'],
-                       pkg_info['version'])
-                    )
+                        '; '.join(uncommon_licenses),
+                        pkg_info['name'],
+                        pkg_info['version'])
+                )
                 sys.exit(1)
 
         yield pkg_info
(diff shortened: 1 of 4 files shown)


piplicenses.py
```diff
@@ -242,23 +242,30 @@ def get_pkg_info(pkg):
...
-
-        if allow_only_licenses and license_name not in allow_only_licenses:
-            sys.stderr.write("license {} not in allow-only licenses was found"
-                             " for package {}:{}".format(
-                                license_name,
-                                pkg_info['name'],
-                                pkg_info['version'])
-                             )
-            sys.exit(1)
+        license_names = set(license_name.split(', '))
```
raimon49: Thanks for your P-R.

I have one concern.

Multiple licenses are separated by `, ` because that is how pip-licenses behaves when it searches and finds them.

There are some "license name" in the world that contain `, `.

See also)
* [SPDX License List | Software Package Data Exchange (SPDX)](https://spdx.org/licenses/)
* [Classifiers · PyPI](https://pypi.org/classifiers/)

What do you think?

Your implementation using `Set.difference()` is simple and elegant. I think there is a benefit to merging these.

raimon49: I would also like comments from recent contributors if they have any.

ping @dkarp0 @cdce8p

test_piplicenses.py
```diff
@@ -599,7 +599,7 @@ def write(self, p):
         self.printed = p
 
 
-def test_output_file_sccess(monkeypatch):
+def test_output_file_success(monkeypatch):
```
raimon49, line 602: Thank you! Nice typo correction 👍

cdce8p (commented):
I'm not sure this change will work as intended. It implicitly assumes that `,` equals `AND` which isn't a given. Take `keyring` for example: As best as I can tell it is dual licensed so it can be distributed together with Python. So as long as you allow / disallow only one of the licenses, it's still fine.

I believe this can only be improved once PEP 639 is accepted. That will specify license expression syntax. Sadly progress on it has more or less stalled recently. https://www.python.org/dev/peps/pep-0639/

piplicenses.py
khatkar: @raimon49 I wasn't aware of licenses containing a comma. To tackle that,  what we can do is use the ; delimiter to concatenate multiple licenses.
For example, two licenses for a package like this;
 - OGC Software License, Version 1.0
 - MIT License

Can be combined as string `"OGC Software License, Version 1.0;MIT License"` and then creating a set from this string based on ; delimiter will result in `set("OGC Software License, Version 1.0", "MIT License")` and this will work as intended.

dkarp0:
This would be a great contribution! If you're allowing either of the licenses then I don't think that will work with your current implementation using `set.difference()` though. Consider the following:

```
keyring     21.4.0     Python Software Foundation License, MIT License
fastly        0.5.1        MIT License

pip-licenses --allow-only="Python Software Foundation License" - Fail
pip-licenses --allow-only="MIT License" - Succeed
pip-licenses --allow-only="Python Software Foundation License;MIT License" - Succeed
```

khatkar:
@dkarp0 Yes Daniel, the current implementation checks for all. I was proposing the idea, and didn't made change until now. It should work with the new commit.

khatkar:
This PR is ready for another review. I've addressed concerns and tests pass. Let me know if I should squash into a single commit.

raimon49:
@khatkar You don't have to squash commits into one.

raimon49 (approved):
I also prefer Daniel's suggestion of a behavior that works as an `OR`.

I will figure out how to describe in the CHANGELOG that the output result of multiple licenses will be slightly different before I release the next version. This is a small change. The next version number will probably be `3.5.0`.

Thank you, great contributors! @khatkar @dkarp0 @cdce8p

Issue #113, opened 2021-10-11
 pip-21.3 breaks pip-licenses


stishkin:
/mnt/user$ pip-licenses -uf json
```
Traceback (most recent call last):
  File "/home/user/.local/lib/python3.8/site-packages/piplicenses.py", line 42, in <module>
    from pip._internal.utils.misc import get_installed_distributions
ImportError: cannot import name 'get_installed_distributions' from 'pip._internal.utils.misc' (/home/user/.local/lib/python3.8/site-packages/pip/_internal/utils/misc.py)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/user/.local/bin/pip-licenses", line 5, in <module>
    from piplicenses import main
  File "/home/user/.local/lib/python3.8/site-packages/piplicenses.py", line 44, in <module>
    from pip import get_installed_distributions
ImportError: cannot import name 'get_installed_distributions' from 'pip' (/home/user/.local/lib/python3.8/site-packages/pip/__init__.py)
```

SlavaSkvortsov:
If someone still wants to have the latest pip version you can use a fork while my PR is still on the review - `git+git://github.com/adjust/pip-licenses.git@v-3.5.3`

PR #114, opened 2021-10-12
#113 Support latest pip version

SlavaSkvortsov:
Fixes issue #113.
In pip==21.3 `get_installed_distributions` function [was removed](https://github.com/pypa/pip/commit/d051a00fc57037104fca85ad8ebf2cdbd1e32d24). I copied this function from pip with removing unused params

I think the best solution would be migrating the whole library to [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html) (it has a [backport](https://pypi.org/project/importlib-metadata/)) because pip [deprecates](https://mail.python.org/pipermail/distutils-sig/2017-October/031642.html) usage of their internal API. But as a hotfix, we can use this solution.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -45,8 +45,12 @@ except ImportError:  # pragma: no cover
         from pip import get_installed_distributions
     except ImportError:
         def get_installed_distributions():
-            from pip._internal.metadata import get_default_environment, get_environment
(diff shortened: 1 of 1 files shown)


raimon49:
@SlavaSkvortsov Thanks for the great contribution!

The matrix test has been added, which is good 👍

PR #116, opened 2021-10-14
reference using setuptools instead of pip internal

tejaschumbalkar:
Issue: #113 

Reference function using setuptools instead of pip internal function. Reference: https://github.com/pypa/pip/issues/5243

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -37,31 +37,7 @@ from email.parser import FeedParser
 from enum import Enum, auto
 from functools import partial
 from typing import List, Optional, Sequence, Text
-
-try:
-    from pip._internal.utils.misc import get_installed_distributions
-except ImportError:  # pragma: no cover
-    try:
-        from pip import get_installed_distributions
-    except ImportError:
-        def get_installed_distributions():
-            from pip._internal.metadata import (
-                get_default_environment, get_environment,
-            )
-            from pip._internal.metadata.pkg_resources import (
-                Distribution as _Dist,
-            )
-            from pip._internal.utils.compat import stdlib_pkgs
-
-            env = get_default_environment()
-            dists = env.iter_installed_distributions(
-                local_only=True,
-                skip=stdlib_pkgs,
-                include_editables=True,
-                editables_only=False,
-                user_only=False,
-            )
-            return [dist._dist for dist in dists]
+import pkg_resources
 
 from prettytable import PrettyTable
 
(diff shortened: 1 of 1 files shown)


piplicenses.py
```diff
@@ -37,31 +37,7 @@
...
-            env = get_default_environment()
-            dists = env.iter_installed_distributions(
-                local_only=True,
-                skip=stdlib_pkgs,
-                include_editables=True,
-                editables_only=False,
-                user_only=False,
-            )
-            return [dist._dist for dist in dists]
+import pkg_resources
```
raimon49, line 40: I understood that it is better not to import functions from `pip._internal`. However, this ported code seems to have the advantage of working in many user environments.

Will `import pkg_resources` succeed in all user environments? For example, it seems to exist in `pip._vendor.pkg_resources`.

jpmn:
Any hope in seeing a fix for this issue merged in? Using `pip=21.3.1`, it throws an exception:

```
File ".venv\Scripts\pip-licenses.exe\__main__.py", line 4, in <module>
File ".venv\lib\site-packages\piplicenses.py", line 44, in <module>
  from pip import get_installed_distributions
ImportError: cannot import name 'get_installed_distributions' from 'pip' (.venv\lib\site-packages\pip\__init__.py)
```

raimon49:
@jpmn I haven't decided if I want to merge this PR or not.

Please install the latest version of pip-licenses in your environment and see if it solves the error.

```bash
# Update pip-licenses version to 3.5.3 or later
$ pip install -U pip-licenses
```

jpmn:
I am using the latest version of `pip-licenses`, but I had to freeze my `pip` version until there is a fix, which isn't good. I might consider dropping the library in the medium-term if it is blocking my upgrade of `pip`.

Don't you experience the same behavior with more recent versions of `pip`?

worroc:
This fix is required for many users, it is not acceptable to pin the pip version for a such long time. Is there road map to support pip >= 22?

raimon49:
@worroc Who are "many users"? If this is about you, we need information about your environment.

At least in my environment it is working with pip 22.x.
```bash
$ pip-licenses --with-system
 Name               Version   License
 PTable             0.9.2     BSD License
 botocore           1.21.53   Apache Software License
 jmespath           0.10.0    MIT License
 pip                22.2.2    MIT License
 pip-licenses       3.5.3     MIT License
```

raimon49:
This proposal was superseded by another excellent proposal that uses importlib\_metadata.

Issue #128, opened 2022-10-25
pip-licenses broken in python 3.11


b-kamphorst:
`pip-licenses` is broken in python 3.11 for `pip>=22.1`.

The issue is roughly the same as in #113. Log:

```bash
$ pip-licenses
Traceback (most recent call last):
  File "/builds/gitlab/.../.venv/bin/pip-licenses", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/builds/gitlab/.../stubs/.venv/lib/python3.11/site-packages/piplicenses.py", line 900, in main
    output_string = create_output_string(args)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/gitlab/.../stubs/.venv/lib/python3.11/site-packages/piplicenses.py", line 558, in create_output_string
    table = create_licenses_table(args, output_fields)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/builds/gitlab/.../stubs/.venv/lib/python3.11/site-packages/piplicenses.py", line 303, in create_licenses_table
    for pkg in get_packages(args):
  File "/builds/gitlab/.../stubs/.venv/lib/python3.11/site-packages/piplicenses.py", line 2[54](...), in get_packages
    pkg_name = pkg.project_name
               ^^^^^^^^^^^^^^^^
AttributeError: 'PathDistribution' object has no attribute 'project_name'
```

nejch:
@raimon49 I just opened this issue and your comment popped up a few seconds after :)

I installed the rc1 and this works for me on 3.11.

raimon49:
RC2 is now available following bug reports and patches. #132 

```bash
pip install 'pip-licenses==4.0.0rc2'
```

We continue to welcome feedback.

raimon49:
pip-licenses 4.0.0 has been officially released. Thank you all for your contributions.
https://pypi.org/project/pip-licenses/4.0.0/

PR #129, opened 2022-10-25
feat: depend on importlib_metadata rather than pip

b-kamphorst:
Especially, rather than functions that are internal to pip.

Solves https://github.com/raimon49/pip-licenses/issues/128.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+        python-version: ['3.8', '3.9', '3.10', '3.11']
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -13,7 +13,6 @@ jobs:
     strategy:
       matrix:
         python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
-        pip-version: ['', '==21.2.4']
 
     steps:
     - uses: actions/checkout@v2
@@ -23,7 +22,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip${{ matrix.pip-version }}
+        python -m pip install --upgrade pip
         if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
     - name: Test with pytest
       run: |
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
         pip-version: ['', '==21.2.4']
 
     steps:
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -50,10 +50,7 @@ except ImportError:  # pragma: no cover
     from prettytable import NONE as RULE_NONE
     PTABLE = False
 
-if sys.version_info < (3, 8):
-    import importlib_metadata
-else:
-    from importlib import metadata as importlib_metadata
+from importlib import metadata as importlib_metadata
 
 open = open  # allow monkey patching
 
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -138,7 +138,10 @@ def get_packages(args: "CustomNamespace"):
 
         pkg_files = pkg.files or ()
         pattern = re.compile(file_names_rgx)
-        matched_rel_paths = filter(lambda file: pattern.match(file.name), pkg_files)
+        matched_rel_paths = filter(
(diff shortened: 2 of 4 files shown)


b-kamphorst:
@raimon49 I look forward to your thoughts! If you approve, please let me know what version you'd like to assign so I can update that accordingly.

b-kamphorst:
Note that, for python 3.7, the proposed change implies that `importlib_metadata` (Apache license) is installed as a `pip-licenses` dependency. I think this dependency should be added to `SYSTEM_PACKAGES` just like `PTable`/`prettytable` ?

raimon49:
@b-kamphorst Thanks, I have installed Python 3.11.0 in my environment as well and confirmed the error you reported.

piplicenses.py
```diff
@@ -78,6 +50,11 @@ def get_installed_distributions():
     from prettytable import NONE as RULE_NONE
     PTABLE = False
 
+if sys.version_info < (3, 8):
+    import importlib_metadata
+else:
+    from importlib import metadata as importlib_metadata
+
```
raimon49: > Note that, for python 3.7, the proposed change implies that `importlib_metadata` (Apache license) is installed as a `pip-licenses` dependency. I think this dependency should be added to `SYSTEM_PACKAGES` just like `PTable`/`prettytable` ?

We are currently trying to stop using `PTable` and move to `prettytable`. #82

> In the current version of prettytable 2.5.0, the number of dependent packages has increased.
> 
> https://github.com/jazzband/prettytable/blob/2.5.0/setup.cfg
> ```
> [options]
> install_requires =
>     wcwidth
>     importlib-metadata;python_version < '3.8'
> ```

To complete the migration, we thought we would add `wcwidth` / `importlib-metadata` to `SYSTEM_PACKAGES`.

Alternatively, we could choose to upgrade as pip-licenses 4.0 (allow broken changes), stop supporting Python 3.7, and not depend on `importlib-metadata`.

@b-kamphorst @johnthagen I would like to know if you have an opinion.

johnthagen: Given the alternatives, I would vote to keep things simple and drop support for 3.7.

This line will help `pip` install an older version of `pip-licenses` for those still using 3.7 for a short time:

https://github.com/raimon49/pip-licenses/blob/12d5d5f25d6347ec671b723f548a43a87d84c7fd/setup.cfg#L23

.github/workflows/python-package.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
         pip-version: ['', '==21.2.4']
```
raimon49: `pip-version` matrix setting was introduced to guarantee the operation of multiple pip versions.

Your contribution will make this setup unnecessary.

We would like to see the YAML configuration file changed as follows

As-Is:
```yaml
    strategy:
      matrix:
        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
        pip-version: ['', '==21.2.4']
# snip
    - name: Install dependencies
      run: |
        python -m pip install --upgrade pip${{ matrix.pip-version }}
```

To-Be:
```yaml
    strategy:
      matrix:
        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
# snip
    - name: Install dependencies
      run: |
```

piplicenses.py
b-kamphorst: I'm not a fan per se -- python 3.7 is still [officially supported](https://www.python.org/downloads/) for 8 more months. Having said that, given the stability of `pip-licenses` over the last years, I'd like to think that the current version will suffice for those 8 months.

Just to be sure: shouldn't we update the line `python_requires = ~=3.7` to `python_requires = ~=3.8` and drop the classifier for Python 3.7 if we want pip to do the right thing (e.g. install `pip-licenses==3.5.5` instead of `4.0.0`)?

b-kamphorst: I thought of alternatives (e.g. detect whether `pip-licenses`'s dependencies are also dependencies of other packages), but there probably are cases for which this does not work. As such, I currently don't have better ways to offer than what is already considered.

johnthagen: > Just to be sure: shouldn't we update the line python_requires = ~=3.7 to python_requires = ~=3.8 and drop the classifier for Python 3.7 if we want pip to do the right thing 

Yes, exactly.

raimon49:
@b-kamphorst Thanks for the good discussion and great contribution!

I have created a branch for the v4.0.0 release (same commit as the master branch) for work on the following tasks. Can you please change the base branch of this Pull Request to `release-4.0.0` ?

Merge this Pull Request into the `release-4.0.0` and then move on to the next task.

- Migrate dependency library from `PTable` to `prettytable`
- Bump up version to `4.0.0rc1`
  - Upload the release candidate version to PyPI because it involves breaking changes
  - We will officially release it after getting feedback from users
- Update README documentation and more

b-kamphorst:
> @b-kamphorst Thanks for the good discussion and great contribution!

My pleasure!

> I have created a branch for the v4.0.0 release (same commit as the master branch) for work on the following tasks. Can you please change the base branch of this Pull Request to `release-4.0.0` ?

Done. With that I think my work here is done? Also, #116 can be closed.

PR #137, opened 2022-11-25
Add type annotations

b-kamphorst:
Hi @raimon49, this PR looks intimidating so please have a look at the commits one-by-one. For example, just running a linter already caused a huge diff. I'm not entirely happy with the type hints but this is as far as I could go without actually changing the logic. 

I think type annotations would have helped me a lot for my first contribution, and it will also make it easier to make future contributions. Of course that is subjective. I can imagine that you're not quite looking for a PR like this, if so I'll just close the PR. Looking forward to your thoughts.


raimon49:
@b-kamphorst Thanks, this diff looks good to me, I am glad that you have shown the utmost respect for this Git repo style.

It's a good idea to introduce a code formatter. This PR would be acceptable.

.github/workflows/python-package.yml
```diff
@@ -2,39 +2,46 @@ name: Python package
...
-        fail_ci_if_error: true
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Lint with black
+        uses: psf/black@stable
+        with:
+          options: "--check --diff --line-length=79"
```
raimon49, line 28: A make target to run Lint and formatting would seem to be better.

For example, define `make lint` in the Makefile in the same way as `make format`. Of course, run by IDEs would be modern. However, I often just use make.

b-kamphorst: I've just added a `make lint` command in favor of `make format` (I suppose one always wants to perform all linting). I don't see a conflict with having both the `make` command and the github actions step, as the second one validates linting for all commits (also by those people that didn't look at the Makefile, such as myself). If you agree, feel free to merge. If not, just let me know what to change!

raimon49 (approved):
Thank you @b-kamphorst ! . I will merge this code and v4.0.2 will be released. There is no change in usage from users, but it is a valuable milestone.

PR #139, opened 2022-12-08
Escape unicode output (to e.g. &#123;) in the html output

timw6n:
This should fix broken author names etc. when viewing the output in browser.

My current workflow is to save the `--format html` output from pip-licenses straight to a file. To have non-broken author names when viewing that in a browser, need to either wrap the output in a full html document that sets an encoding or, as I'm doing now, pipe through `python -c "import sys; s = sys.stdin.read(); print(s.encode('ascii', errors='xmlcharrefreplace').decode('ascii'))"` which is basically the change in this PR.

It feels better to add this here, rather than upstream in prettytable, since that package works fine outputting utf-8 in the context of including tables in larger documents.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -540,7 +540,8 @@ def create_output_string(args: CustomNamespace) -> str:
     sortby = get_sortby(args)
 
     if args.format_ == FormatArg.HTML:
-        return table.get_html_string(fields=output_fields, sortby=sortby)
+        html = table.get_html_string(fields=output_fields, sortby=sortby)
+        return html.encode('ascii', errors='xmlcharrefreplace').decode('ascii')
     else:
         return table.get_string(fields=output_fields, sortby=sortby)
 
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -541,7 +541,7 @@ def create_output_string(args: CustomNamespace) -> str:
 
     if args.format_ == FormatArg.HTML:
         html = table.get_html_string(fields=output_fields, sortby=sortby)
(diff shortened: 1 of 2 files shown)


piplicenses.py
```diff
@@ -540,7 +540,8 @@ def create_output_string(args: CustomNamespace) -> str:
     sortby = get_sortby(args)
 
     if args.format_ == FormatArg.HTML:
-        return table.get_html_string(fields=output_fields, sortby=sortby)
+        html = table.get_html_string(fields=output_fields, sortby=sortby)
+        return html.encode('ascii', errors='xmlcharrefreplace').decode('ascii')
```
raimon49: Thanks for the PR. Looks good.

Linter was introduced in CI some time ago.

The following code styles are noted could you please correct the string? or `make lint` will also fix it.

```diff
+++ piplicenses.py	2022-12-08 11:34:58.939191 +0000
@@ -539,11 +539,11 @@
 
     sortby = get_sortby(args)
 
     if args.format_ == FormatArg.HTML:
         html = table.get_html_string(fields=output_fields, sortby=sortby)
-        return html.encode('ascii', errors='xmlcharrefreplace').decode('ascii')
+        return html.encode("ascii", errors="xmlcharrefreplace").decode("ascii")
```

timw6n: Done. Apologies, hadn't realised that was in place.

raimon49 (approved):
Thank you for your contribution!

This PR will be merged and released as v4.0.3. Hope this helps your workflow.

raimon49:
v4.0.3 is available.
https://pypi.org/project/pip-licenses/4.0.3/

Issue #145, opened 2023-02-23
Ignore case when matching license names


khatkar:
Does it make sense to match license names ignoring the case?
For the flags `--allow-only` and `--fail-on`, currently the user has to provide all case combinations which feels redundant.
For example,
```bash
$ pip-licenses
Name      Version   License
PIMS      0.5       BSD-3-clause
protobuf  3.20.1    BSD-3-Clause
```

Now running `pip-licenses --allow-only "BSD-3-Clause"` will fail for the pims package.
So the user has to set `--allow-only "BSD-3-Clause;BSD-3-clause"`. If the user maintains a list of allow-only or fail-only licenses as part of their code, long term stability of such code is affected as new case combinations come up. It'd be nice if we can support case-insensitive matching.

PR #146, opened 2023-02-24
Case-insensitive license name matching

khatkar:
* Matching license names with `allow-only` and `fail-on` options is no longer case-sensitive #145 
* Updated Readme and tests

### DIFF
--- a/test_piplicenses.py
+++ b/test_piplicenses.py
@@ -29,6 +29,8 @@ from piplicenses import (
     CompatibleArgumentParser,
     FromArg,
     __pkgname__,
+    case_insensitive_set_diff,
+    case_insensitive_set_intersect,
     create_licenses_table,
     create_output_string,
     create_parser,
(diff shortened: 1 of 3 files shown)


test_piplicenses.py
```diff
@@ -751,7 +751,7 @@ def test_allow_only(monkeypatch) -> None:
 
 
 def test_fail_on(monkeypatch) -> None:
-    licenses = ("MIT License",)
+    licenses = ("MIT license",)
```
raimon49, line 784: Indeed, this will ensure that `case_insensitive_set_intersect` is working.

But both `case_insensitive_set_intersect` and `case_insensitive_set_diff` are simple functions. Could we add a stand-alone test case in terms of expressing these specs?

For example, there is a similarly simple function called `output_colored`, which has the following independent test cases:

* `test_output_colored_normal`
* `test_output_colored_bold`

khatkar: Sure, I have added two independent tests cases in the new commit.

raimon49: Looks good. Thanks.

raimon49 (approved):
A release including this PR will be available on PyPI as v4.1.0

PR #147, opened 2023-03-02
Release 4.1.0

### LINKED ISSUE #145: Ignore case when matching license names

raimon49:
This release branch is fix #145

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## CHANGELOG
 
+### 4.1.0
+
+* Support case-insensitive license name matching around `--fail-on` and `--allow-only` parameters
+
 ### 4.0.3
 
 * Escape unicode output (to e.g. `&#123;`) in the html output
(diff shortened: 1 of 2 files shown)


raimon49:
Just shipped.
https://pypi.org/project/pip-licenses/4.1.0/

PR #150, opened 2023-03-21
Add `--python` option

### LINKED ISSUE #149: Retrieve licenses from other environment
Somehow related to #107 

Currently, we can only get packages from the environments we are running pip-licenses on. This is a problem for two reasons

- As said in mentioned issue, we cannot use it as a pre-commit hook easily. There is the possibility to use a hook within the activated environment, but it's clearly not following its guidelines
- Using this libraries implies installing it in your venv, and thus contaminating it by installing packages you wouldn't need otherwise. It's no great given the purpose of this lib is to identify licenses of installed packages

I think there is a simple fix to this.

Indeed, thepckages are discovered via `importlib` `distributions()` function. As it can be seen here : https://github.com/raimon49/pip-licenses/blob/master/piplicenses.py#L193

By looking at the documentation and code of importlib, it seems to me that `distributions` should accept a `path`  argument with a list of folders to search into.
See
https://importlib-metadata.readthedocs.io/en/latest/api.html#importlib_metadata.distributions

and https://github.com/python/importlib_metadata/blob/700f2c7d74543e3695163d5487155b92e6f04d65/importlib_metadata/__init__.py#L809
 
`distributions(**kwargs)` calls `Distribution.discover(**kwargs)` which expects either a `Context` object or its args for creating one (see https://github.com/python/importlib_metadata/blob/700f2c7d74543e3695163d5487155b92e6f04d65/importlib_metadata/__init__.py#L387 )

All in all, calling `distributions([folder_path])` will retrieve all packages metadata in the given `folder_path` which could be a CLI parameter.

What do you think ?

ClementPinard:
Fixes #149
Still Work in Progress, happy to receive feedback !

Not particularly happy with the fact that you have to specify until the `site-packages` folder and not just the venv folder. suggestions welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,6 +97,21 @@ Execute the command with your venv (or virtualenv) environment.
 
 ### Common options
 
+#### Option: path
+
+By default, this tools finds the packages from the environment pip-licenses is launched from, by searching in `sys.path` folders. In the case you want to search for packages in an other environment (e.g. if you want to run pip-licenses from its own isolated environment), you can specify a list of folders where to search from.
+
+```bash
+(venv) $ pip-licenses --with-system | grep pip
+ pip                       22.3.1       MIT License
+ pip-licenses              4.1.0        MIT License
+```
+
+```bash
+(venv) $ pip-licenses --path=</path/to/other/env>/lib/<python_version>/site-packages --with-system | grep pip
+ pip                       23.0.1       MIT License 
+```
+
 #### Option: from
 
 By default, this tool finds the license from [Trove Classifiers](https://pypi.org/classifiers/) or package Metadata. Some Python packages declare their license only in Trove Classifiers.
--- a/README.md
+++ b/README.md
@@ -97,9 +97,9 @@ Execute the command with your venv (or virtualenv) environment.
 
 ### Common options
 
-#### Option: path
+#### Option: python
 
-By default, this tools finds the packages from the environment pip-licenses is launched from, by searching in `sys.path` folders. In the case you want to search for packages in an other environment (e.g. if you want to run pip-licenses from its own isolated environment), you can specify a list of folders where to search from.
+By default, this tools finds the packages from the environment pip-licenses is launched from, by searching in current python's `sys.path` folders. In the case you want to search for packages in an other environment (e.g. if you want to run pip-licenses from its own isolated environment), you can specify a path to a python executable. The packages will be searched for in the given python's `sys.path`, free of pip-licenses dependencies.
 
 ```bash
 (venv) $ pip-licenses --with-system | grep pip
@@ -108,7 +108,7 @@ By default, this tools finds the packages from the environment pip-licenses is l
 ```
 
 ```bash
-(venv) $ pip-licenses --path=</path/to/other/env>/lib/<python_version>/site-packages --with-system | grep pip
+(venv) $ pip-licenses --python=</path/to/other/env>/bin/python --with-system | grep pip
  pip                       23.0.1       MIT License 
 ```
 
--- a/README.md
+++ b/README.md
@@ -11,6 +11,7 @@ Dump the software license list of Python packages installed with pip.
 * [Usage](#usage)
 * [Command\-Line Options](#command-line-options)
     * [Common options](#common-options)
+        * [Option: python](#option-python)
         * [Option: from](#option-from)
         * [Option: order](#option-order)
         * [Option: format](#option-format)
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -194,7 +194,7 @@ def get_packages(
     def get_python_sys_path(executable: str) -> list[str]:
         script = "import sys; print(' '.join(filter(bool, sys.path)))"
         output = subprocess.run(
-            [executable, "-c", script], capture_output=True
+            [executable, "-c", script], capture_output=True, env={}
         )
         return output.stdout.decode().strip().split()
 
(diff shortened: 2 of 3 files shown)


ClementPinard:
here is how you can use it if you want licenses (and version numbers) from another environment. This would be particularly useful in a CI where you want exact license and versions of your project dependencies, which could be clashing with pip-licenses's dependencies (even though they are minimal)

```bash
source /path/to/venv/bin/activate
/path/to/pip-licenses --path $(python -c "import sys; print(','.join(filter(bool, sys.path)))")
```

this would be particularly interesting with pip-licenses installed in you system python

example with poetry
```bash
pip install pip-licences
source `poetry env info --path`/bin/activate
pip-licenses --path $(python -c "import sys; print(','.join(filter(bool, sys.path)))")
```

thejcannon:
> Not particularly happy with the fact that you have to specify until the site-packages folder and not just the venv folder. 

I've seen other tools allow for a `--python` flag pointing to a python interpreter, then it grabs `sys.path` from that. So you'd point to `<venv>/bin/python` which is much more stomachable and also helps with custom Python shenanigans and might handle `.pth` files better.

ClementPinard:
So I guess launching another python from subprocess ? and then getting back the printed sys.path ? That would be possible indeed ! I'll try to come up with something that uses both

raimon49:
I am not sure if I should merge this PR and include it in the next release 4.2.0.

Is the implementation of the `--path` option complete?

ClementPinard:
Hey,
sorry for late answer. If no objections, I'm going to implement the `--python`  option as suggested by @thejcannon very soon (today if I find the time)

From that we can work on potential review comments, and I think the feature will be complete.

ClementPinard:
So I said I would keep both `--path` and `--python`. Turns out, when I had to give an example of how you would need a path option, I could not come up with something useful not already done by the `--python` .

I you need to get licenses from an extra folder, you might as well install the package with `-e` option instead of playing with `sys.path`

As a result, I decided to only keep the `--python` option, to not add a superfluous option in an already packed arg parser. Feel free to share your thoughts on this, and an example on how you would need it in a clean way :upside_down_face: 

@raimon49 I now think this PR is finished, and open for review.

README.md
```diff
@@ -97,6 +97,21 @@ Execute the command with your venv (or virtualenv) environment.
 
 ### Common options
 
+#### Option: python
+
+By default, this tools finds the packages from the environment pip-licenses is launched from, by searching in current python's `sys.path` folders. In the case you want to search for packages in an other environment (e.g. if you want to run pip-licenses from its own isolated environment), you can specify a path to a python executable. The packages will be searched for in the given python's `sys.path`, free of pip-licenses dependencies.
```
raimon49, line 103: I also checked the importlib_metadata source code. It was using `sys.path` by default as you described 👍 

There is a ToC at the top of the README document. Could you also add a link to "Option: python"?

piplicenses.py
```diff
@@ -190,7 +191,19 @@ def filter_string(item: str) -> str:
...
+            [executable, "-c", script], capture_output=True
+        )
+        return output.stdout.decode().strip().split()
+
+    if args.python == sys.executable:
+        search_paths = sys.path
+    else:
+        search_paths = get_python_sys_path(args.python)
+
+    pkgs = importlib_metadata.distributions(path=search_paths)
```
raimon49, line 209: To resolve the points that Codecov points out, we need a test case that mocks the call to `importlib_metadata.distributions()`, for example. Any ideas on how to do this?

Of course, it is preferable to have test cases, but if it is too difficult, I will merge the PR as is.

ClementPinard: Interesting to that L203 is not covered, I would have thought that the L202 would not be covered since we don't test other python environments

ClementPinard:
I add a test that creates a temporary empty virtual env, and thus should only find pip and setuptools.

I had some troubles figuring out why it still found all packages in main virtual environment, and solved this problem by calling the alternative python executable with empty environment variables for `PYTHONPATH` and `VIRTUAL_ENV`

```python
output = subprocess.run(
            [executable, "-c", script],
            capture_output=True,
            env={**os.environ, "PYTHONPATH": "", "VIRTUAL_ENV": ""},
        )
```

Would there be some usecase where keeping them is actually important ? FYI, not resetting `PYTHONPATH` breaks the tests because it is set to the main virtual environment, resetting `VIRTUAL_ENV` does not

raimon49:
> I add a test that creates a temporary empty virtual env, and thus should only find pip and setuptools.

Oh, it had never occurred to me that this was the way to do it. I consider it a good enough test case.

Why are CI tests failing only in the Python 3.8 environment?

ClementPinard:
Looks like I was not typehinting the right way. The venvBuilder context is a `types.SimpleNamespace` object (https://docs.python.org/3.8/library/types.html#types.SimpleNamespace), but I typehinted it as `venv.SimpleNamespace`  I suspect `venv` did import `SimpleNamespace` but only after a certain version, because mypy did not complain for my python3.10 version

I put back the original type in the typehinbt and we should be good to go.  If not, I'll just remove the typehint altogether.

raimon49:
@ClementPinard Thank you for your great contribution! I will merge this PR into the next release candidate version.

raimon49:
pip-licenses version 4.2.0 RC2 is now available. This PR includes.

If you have any concerns about the actual use of the system, please provide feedback. Thanks.
```bash
$ pip install "pip-licenses==4.2.0rc2"
```

PR #153, opened 2023-03-24
feat: Use `author-email` metadata as fallback if `author` is None

g3n35i5:
I noticed that some packages (e.g. [termcolor](https://github.com/termcolor/termcolor)) do not declare an author (metadata field `author`) but the email address of the author (metadata field `author-email`). For this reason I think it is appropriate to use this value as a fallback value.

Current output:
```
Name       Version  License      Author                                      
termcolor  2.2.0    MIT License  UNKNOWN
```

New output:
```
Name       Version  License      Author                                      
termcolor  2.2.0    MIT License  Konstantin Lepa <konstantin.lepa@gmail.com>
```

Other examples for packages with defined `author-email` but no `author`
* https://pypi.org/project/tomli/
* https://pypi.org/project/tabulate/
* https://pypi.org/project/uvicorn/

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -93,7 +93,7 @@ SUMMARY_OUTPUT_FIELDS = (
     "License",
 )
 
-METADATA_KEYS: dict[str, list[str]] = {
+METADATA_KEYS = {
     "home-page": ["home-page"],
     "author": ["author", "author-email"],
     "license": ["license"],
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -93,13 +93,12 @@ SUMMARY_OUTPUT_FIELDS = (
     "License",
 )
 
-
-METADATA_KEYS = (
-    "home-page",
-    "author",
-    "license",
-    "summary",
-)
+METADATA_KEYS: dict[str, tuple[str, ...]] = {
+    "home-page": "home-page",
+    "author": ["author", "author-email"],
+    "license": "license",
+    "summary": "summary",
+}
 
 # Mapping of FIELD_NAMES to METADATA_KEYS where they differ by more than case
 FIELDS_TO_METADATA_KEYS = {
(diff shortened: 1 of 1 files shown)


piplicenses.py
```diff
@@ -93,13 +93,12 @@
...
 )
 
-
-METADATA_KEYS = (
-    "home-page",
-    "author",
-    "license",
-    "summary",
-)
+METADATA_KEYS: dict[str, list[str]] = {
```
g3n35i5:
```suggestion
METADATA_KEYS = {
```

g3n35i5:
Build failure on python 3.8 seems unrelated
```
[2023-03-30T10:13:45.738Z] ['error'] There was an error running the uploader: Error uploading to [https://codecov.io:](https://codecov.io/) Error: There was an error fetching the storage URL during POST: 404 - {'detail': ErrorDetail(string='Unable to locate build via Github Actions API. Please upload with the Codecov repository upload token to resolve issue.', code='not_found')}
```

raimon49:
I too was concerned about the CI error, but it appears that it is now being reported in large numbers by the Codecov service.

@g3n35i5 There appears to be no problem with your fallback code. I will merge this patch and release the next version to PyPI.

Thanks!!

piplicenses.py
```diff
@@ -93,13 +93,12 @@
...
-    "author",
-    "license",
-    "summary",
-)
+METADATA_KEYS = {
+    "home-page": ["home-page"],
+    "author": ["author", "author-email"],
+    "license": ["license"],
+    "summary": ["summary"],
+}
```
raimon49, line 101: I understand that this is unavoidable in order to handle `"author-email"` well, but this structure is redundant.

I have no idea how to write smarter either...

g3n35i5: I could implement a `Union[str, list[str]]` approach with the following logic

```python
for field_name, field_selectors in METADATA_KEYS.items():
    value = None
    if isinstance(field_selectors, list):
        for selector in field_selectors:
            value = metadata.get(selector, None)
                if value:
                    break
    else:
        value = metadata.get(field_selectors, None)
    pkg_info[field_name] = value or LICENSE_UNKNOWN
```

but i think this makes it even more fuzzy :laughing:

raimon49: Hm... yes, indeed 😅

raimon49:
@g3n35i5 Thanks for your contribution.

The next release candidate version has been uploaded to PyPI.
https://pypi.org/project/pip-licenses/4.2.0rc1/

Install it in your environment and make sure it is working as expected.
```bash
pip install pip-licenses==4.2.0rc1
```

If you have code to fix, please send a PR to the `release-4.2.0` branch.

g3n35i5:
> Install it in your environment and make sure it is working as expected.

Works as expected :)

PR #158, opened 2023-04-20
feat: adding `--no-version` option

dennybiasiolli:
to avoid exporting the version number of packages.

I.E. for security reasons you want to export licenses without the specific version number.


raimon49:
I thought the version number was essential from a software linesense audit perspective.

> I.E. for security reasons you want to export licenses without the specific version number.

Your opinion made me think that indeed there is a demand for this kind of thing.

Thanks for the PR. The implementation and test code look fine. I will merge it later.

raimon49:
@dennybiasiolli We have published a merged version of your contribution to PyPI.
https://pypi.org/project/pip-licenses/4.3.0/

Thanks!

PR #161, opened 2023-05-04
Fix normalize pkg name

### LINKED ISSUE #160: Treat `-` and `_` as equivalent in package names
`pip` treats `-` and `_` as equivalent, but `pip-licenses` differentiates between the two when using the `-p` flag. Having `pip-licenses` also treat `-` and `_` as equivalent makes it easier to work with.

In my particular use case, I have a venv that contains packages from a `requirements.txt` and also additional packages that I don't want licenses of. I then passed the packages in the `requirements.txt` to `pip-licenses -p` and was surprised when the license for `typing-extensions` wasn't found. However, passing `typing_extensions` instead to `pip-licenses` works.

`requirements.txt` in this form can be created from, e.g. `pipenv requirements`.

raimon49:
This PR will fix #160

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## CHANGELOG
 
+### 4.3.1
+
+* Fix to treat package names as normalized as in [PEP 503](https://peps.python.org/pep-0503/) with `--packages` and `--ignore-packages` option
+
 ### 4.3.0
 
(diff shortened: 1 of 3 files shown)


raimon49:
Shipped.
https://pypi.org/project/pip-licenses/4.3.1/

PR #162, opened 2023-05-18
package URL extraction improvements 

imrehg:
This PR has 2 parts:

1. Switch to case-insensitive matching of fallback amongst the entries `Project-URL`. Came across many projects that do use lowercase for their entries, and previously they were listed with URL being `UNKOWN`.
2. Add `repository` as another fallback. This I see popping up quite a bit among projects, instead of `source`.

Happy to hear feedback on either of these. 1. is a blocker for us on some projects, 2. is a potential nice-to-have but optional (added in a separate commit).

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -122,7 +122,13 @@ def extract_homepage(metadata: Message) -> Optional[str]:
         key, value = entry.split(",", 1)
         candidates[key.strip().lower()] = value.strip()
 
-    for priority_key in ["homepage", "source", "changelog", "bug tracker"]:
+    for priority_key in [
+        "homepage",
+        "source",
+        "repository",
+        "changelog",
+        "bug tracker",
+    ]:
         if priority_key in candidates:
             return candidates[priority_key]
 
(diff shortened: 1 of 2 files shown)


raimon49:
Thanks for the PR.

> Happy to hear feedback on either of these. 1. is a blocker for us on some projects, 2. is a potential nice-to-have but optional (added in a separate commit).

Great, as far as the guideline about [Declaring project metadata](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#declaring-project-metadata), it looks like there are projects that have set up `repository` keywords.

```toml
[project.urls]
homepage = "https://example.com"
documentation = "https://readthedocs.org"
repository = "https://github.com/me/spam.git"
changelog = "https://github.com/me/spam/blob/master/CHANGELOG.md"
```

Also, I see no problem with adding `repository` to the search target because `homepage` has priority as expressed in your test case.

raimon49 (approved):
I will upload a release to PyPI containing this PR on weekend.

raimon49:
@imrehg Uploaded as v4.3.2.
https://pypi.org/project/pip-licenses/4.3.2/

Thank you for your contribution!

PR #186, opened 2024-03-26
Enable partial matching on allow-only/fail-on

### LINKED ISSUE #187: Consider enabling partial matching for licenses
Hello,

there are use cases, where partial matching on license name could be beneficial, such as when one would like to match all versions of a license - say BSD or GPL or  GPLv3 vs GPLv3+.

Consider adding a way to enable partial matching against licenses, as a non-default option, to ease maintenance of long licenses allow/block lists.

Thanks!

costasd:
This changeset introduces a way to do partial matching on license names.

Say, to match all BSD licenses for a given --fail-on=BSD.

To do so, it:
* Introduces a new cli boolean parameter (default false), called --partial-match When enabled, enables partial matching.
* Introduces code to mimic slower, partial-match inntersection/diff methods
* Introduces tests to test how this works

Fixes #187

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,7 @@ Dump the software license list of Python packages installed with pip.
     * [Verify options](#verify-options)
         * [Option: fail\-on](#option-fail-on)
         * [Option: allow\-only](#option-allow-only)
+        * [Option: partial\-match](#option-partial-match)
     * [More Information](#more-information)
 * [Dockerfile](#dockerfile)
 * [About UnicodeEncodeError](#about-unicodeencodeerror)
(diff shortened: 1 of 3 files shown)


raimon49:
Thanks for the implementation idea and code! It looks good because it's the default false, so it won't break the CI of existing users.

I would like to see the role of this option written in the README document. This feature appears to be used in combination with `--fail-on` or `--allow-only`. Users would appreciate an explanation in the "Verify options" section of the README. What do you think?

costasd:
oh missed the readme, you're right, will add it there!

raimon49:
@costasd We have the `--partial-match` option available in v4.4.0 🎉 
https://pypi.org/project/pip-licenses/4.4.0/

Thank you for your contribution.

PR #206, opened 2024-07-23
Release with pyproject.toml

raimon49:
This release contains #203 

In addition, Python 3.8 will soon be EOL.
https://devguide.python.org/versions/

Therefore, we would also like to set the lower limit of Python versions supported by pip-licenses to 3.9.


raimon49:
Shipped.
https://pypi.org/project/pip-licenses/5.0.0/

PR #213, opened 2024-10-11
Add support for License-Expression parsing (PEP 639)

cdce8p:
[PEP 639](https://peps.python.org/pep-0639/) has been (provisionally) accepted recently. `hatch` added support[^1] for it early on and so the first packages which store the license metadata inside the `License-Expression` field start popping up. E.g. for [ftfy](https://pypi.org/project/ftfy/) it looks like this
```
...
License-Expression: Apache-2.0
License-File: LICENSE.txt
...
```

This PR adds the basic parsing support without any validation similar to the existing behavior for the `License` metadata field or the classifier. It might make sense to add the validation at a later point though.

To keep the changes to a minimum, the behavior of `--from=mixed` is preserved. However:

> [!CAUTION]
> **Breaking change**: The `--from=all` output now includes the `License-Expression` value.

[^1]: https://hatch.pypa.io/1.9/config/metadata/#spdx-expression

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -162,6 +162,9 @@ METADATA_KEYS: Dict[str, List[Callable[[Message], Optional[str]]]] = {
         lambda metadata: metadata.get("maintainer-email"),
     ],
     "license": [lambda metadata: metadata.get("license")],
+    "license_expression": [
+        lambda metadata: metadata.get("license-expression")
+    ],
     "summary": [lambda metadata: metadata.get("summary")],
 }
 
@@ -171,6 +174,7 @@ FIELDS_TO_METADATA_KEYS = {
     "Description": "summary",
     "License-Metadata": "license",
     "License-Classifier": "license_classifier",
+    "License-Expression": "license_expression",
 }
 
 
@@ -312,6 +316,7 @@ def get_packages(
             args.from_,
             cast(List[str], pkg_info["license_classifier"]),
             cast(str, pkg_info["license"]),
+            cast(str, pkg_info["license_expression"]),
         )
 
         if fail_on_licenses:
@@ -374,6 +379,7 @@ def create_licenses_table(
                     args.from_,
                     cast(List[str], pkg["license_classifier"]),
                     cast(str, pkg["license"]),
+                    cast(str, pkg["license_expression"]),
                 )
                 license_str = "; ".join(sorted(license_set))
                 row.append(license_str)
@@ -399,6 +405,7 @@ def create_summary_table(args: CustomNamespace) -> PrettyTable:
                     args.from_,
                     cast(List[str], pkg["license_classifier"]),
                     cast(str, pkg["license"]),
+                    cast(str, pkg["license_expression"]),
                 )
             )
         )
@@ -613,7 +620,10 @@ def find_license_from_classifier(classifiers: list[str]) -> list[str]:
 
 
 def select_license_by_source(
-    from_source: FromArg, license_classifier: list[str], license_meta: str
+    from_source: FromArg,
+    license_classifier: list[str],
+    license_meta: str,
+    license_expression: str,
 ) -> set[str]:
     license_classifier_set = set(license_classifier) or {LICENSE_UNKNOWN}
     if (
@@ -622,8 +632,14 @@ def select_license_by_source(
         and len(license_classifier) > 0
     ):
         return license_classifier_set
-    else:
+    elif (
+        from_source == FromArg.META
+        or from_source == FromArg.MIXED
+        and license_meta != LICENSE_UNKNOWN
+    ):
         return {license_meta}
+    else:
+        return {license_expression}
 
 
 def get_output_fields(args: CustomNamespace) -> list[str]:
@@ -635,6 +651,7 @@ def get_output_fields(args: CustomNamespace) -> list[str]:
     if args.from_ == FromArg.ALL:
         output_fields.append("License-Metadata")
         output_fields.append("License-Classifier")
+        output_fields.append("License-Expression")
     else:
         output_fields.append("License")
 
@@ -839,6 +856,7 @@ class FromArg(NoValueEnum):
     META = M = auto()
     CLASSIFIER = C = auto()
     MIXED = MIX = auto()
+    EXPRESSION = EXPR = auto()
     ALL = auto()
 
 
--- a/test_piplicenses.py
+++ b/test_piplicenses.py
@@ -212,6 +212,20 @@ class TestGetLicenses(CommandLineTestCase):
         license_notation_as_classifier = "MIT License"
         self.assertIn(license_notation_as_classifier, license_columns)
 
+    def test_from_expression(self) -> None:
+        from_args = ["--from=expression"]
+        args = self.parser.parse_args(from_args)
+        output_fields = get_output_fields(args)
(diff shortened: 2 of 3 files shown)


stefan6419846:
Isn't the proposed handling against PEP 639? According to https://packaging.python.org/en/latest/specifications/core-metadata/#license, the following holds true:

> As of Metadata 2.4, `License` and `License-Expression` are mutually exclusive. If both are specified, tools which parse metadata will disregard `License` and PyPI will reject uploads. See [PEP 639](https://peps.python.org/pep-0639/#deprecate-license-field).

With PyPI not allowing both keys to be set and `License-Expression` taking precedence otherwise, wouldn't it make more sense to retrieve `license-expression` as the first variant in `METADATA_KEYS["license"]`? This would avoid the breaking change as well.

cdce8p:
> With PyPI not allowing both keys to be set and `License-Expression` taking precedence otherwise, wouldn't it make more sense to retrieve `license-expression` as the first variant in `METADATA_KEYS["license"]`?

Yeah, you're correct. `License-Expression` should take precedence and all other values should be ignored. My initial thought was to preserve backwards compatibility as much as possible and not change the precedence order on an existing option. In practice it shouldn't make much of a difference as "normally" only one should be defined (either License-Expression or License metadata / classifier). Anyway pushed a new commit to update the logic.

> This would avoid the breaking change as well.

Not quite sure about that one. The breaking change is in the `--from=all` output. It adds a new column, so for users who depend on the exact form it's a breaking change.

test_piplicenses.py
```diff
@@ -212,6 +212,20 @@ def test_from_mixed(self) -> None:
...
+        output_fields = get_output_fields(args)
+        table = create_licenses_table(args, output_fields)
+
+        self.assertIn("License", output_fields)
+
+        license_columns = self._create_license_columns(table, output_fields)
+        license_notation_as_expression = "MIT"
+        # TODO enable assert once a dependency uses 'License-Expression'
+        # TODO (maybe black)
+        # self.assertIn(license_notation_as_expression, license_columns)
```
SMoraisAnsys, line 227: If you update the dependency file with `attrs==25.1.0`, you'll end up with a dependency using license expression (see https://github.com/python-attrs/attrs/releases/tag/25.1.0)

test_piplicenses.py
```diff
@@ -239,6 +259,10 @@ def test_from_all(self) -> None:
             "Apache Software License",
         ):
             self.assertIn(license_name, license_classifier)
+        # TODO enable assert once a dependency uses 'License-Expression'
+        # TODO (maybe black)
+        # for license_name in ("MIT",):
+        #     self.assertIn(license_name, license_expression)
```
SMoraisAnsys, line 265: Same comment as before

SMoraisAnsys (changes requested):
Thanks for this PR ! Could you update the dependency file to add a package using license expression ? I proposed `attrs` but any other package doing it is fine :)

cdce8p:
> Could you update the dependency file to add a package using license expression ? I proposed `attrs` but any other package doing it is fine :)

I'd prefer not to add an additional test dependency just for that. The last `prettytable` release also added `License-Expression`. Once #222 is merged, I can update this PR / or do a followup to enable the asserts.

SMoraisAnsys:
> > Could you update the dependency file to add a package using license expression ? I proposed `attrs` but any other package doing it is fine :)
> 
> I'd prefer not to add an additional test dependency just for that. The last `prettytable` release also added `License-Expression`. Once #222 is merged, I can update this PR / or do a followup to enable the asserts.

Attrs is already a dependency (see the top of the dependency file)
But sure, prettiable is a good target :)

cdce8p:
> Attrs is already a dependency (see the top of the dependency file) But sure, prettiable is a good target :)

It's only a transitive dependency on an old pytest version. Will be removed with #217 it seems. Anyway, I'd like to keep dependency bumps separate. Will update the asserts here, if #217 and / or #222 ever gets merged.

SMoraisAnsys:
@raimon49 Sorry for the direct ping but this package doesn't seem to have much activity since 6 months. Could you have a look at this PR or should we assume the repo to be dead and move with a fork ?

martin-traverse:
Hi - any chance of getting something merged for this? Packages are starting to use the new standard. On our project we've started blocking some updates, but that can only be done for so long. Getting a new release onto PyPI that supports the new standard would be really helpful.

bd-gpasternak:
desperately asking that this get merged so that we don't have to perform surgery to roll our own pip license code.

piplicenses.py
```diff
@@ -162,6 +162,9 @@ def normalize_pkg_name(pkg_name: str) -> str:
         lambda metadata: metadata.get("maintainer-email"),
     ],
     "license": [lambda metadata: metadata.get("license")],
```
arthur-proglove, line 164:
```suggestion
    "license": [
        lambda metadata: metadata.get("license-expression"),
        lambda metadata: metadata.get("license")
    ]
```

isn't it just enough to add a new lambda here?

cdce8p: Technically it would be. However, wouldn't be possible to tell if a package has adopted PEP 639 yet or is still using the old `License` key. That's why I've chosen to use a separate mapping here.

johnthagen:
For those in need of this, consider `pip-licenses-cli`, a maintained fork that is a drop-in replacement that supports PEP 639

- https://github.com/raimon49/pip-licenses/issues/227#issuecomment-2885810102

piplicenses.py
```diff
@@ -613,8 +620,14 @@ def find_license_from_classifier(classifiers: list[str]) -> list[str]:
 
 
 def select_license_by_source(
-    from_source: FromArg, license_classifier: list[str], license_meta: str
+    from_source: FromArg,
+    license_classifier: list[str],
+    license_meta: str,
+    license_expression: str,
 ) -> set[str]:
+    if license_expression != LICENSE_UNKNOWN:
```
reactive-firewall: :broom: NIT: minor bug risk

Please consider:

```suggestion
    if not license_expression or license_expression != LICENSE_UNKNOWN:
```

as in:

```python
    if not license_expression or (license_expression != LICENSE_UNKNOWN):
```

piplicenses.py
reactive-firewall: let's go with @cdce8p's implementation here.

piplicenses.py
```diff
@@ -635,6 +648,7 @@ def get_output_fields(args: CustomNamespace) -> list[str]:
     if args.from_ == FromArg.ALL:
         output_fields.append("License-Metadata")
         output_fields.append("License-Classifier")
+        output_fields.append("License-Expression")
```
reactive-firewall, line 651: > [!IMPORTANT]
> The breaking change will require a project version bump.

I'm fine with the change, as it makes logical sense. With the caveat that the output style change should be reviewed before release. (non-blocker)

test_piplicenses.py
reactive-firewall, line 1: ## :+1: LGTM; (`test_piplicenses.py`)

:ballot_box_with_check: Specifically, the test code is fine.

piplicenses.py
markuspi: Hi @reactive-firewall I don't quite understand the implied bug risk here. Would you mind explaining what input the new statement can catch that the old one couldn't?

PR #214, opened 2024-10-11
Replace tomli with tomllib for Python 3.11+

cdce8p:
`tomllib`[^1] was added as a builtin module in 3.11. It's based on `tomli` and as such fully compatible.

[^1]: https://docs.python.org/3/library/tomllib.html

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -42,12 +42,16 @@ from importlib.metadata import Distribution
 from pathlib import Path
 from typing import TYPE_CHECKING, Iterable, List, Type, cast
 
-import tomli
 from prettytable import ALL as RULE_ALL
 from prettytable import FRAME as RULE_FRAME
 from prettytable import HEADER as RULE_HEADER
 from prettytable import NONE as RULE_NONE
 from prettytable import PrettyTable
+
+if sys.version_info >= (3, 11):
+    import tomllib
+else:
+    import tomli as tomllib
 
 if TYPE_CHECKING:
     from email.message import Message
@@ -174,15 +178,16 @@ FIELDS_TO_METADATA_KEYS = {
 }
 
 
-SYSTEM_PACKAGES = (
+SYSTEM_PACKAGES = [
     __pkgname__,
     "pip",
     "prettytable",
     "wcwidth",
     "setuptools",
-    "tomli",
     "wheel",
-)
+]
+if sys.version_info < (3, 11):
+    SYSTEM_PACKAGES.append("tomli")
 
 LICENSE_UNKNOWN = "UNKNOWN"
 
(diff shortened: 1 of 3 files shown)


piplicenses.py
```diff
@@ -43,13 +43,18 @@
...
 from prettytable import FRAME as RULE_FRAME
 from prettytable import HEADER as RULE_HEADER
 from prettytable import NONE as RULE_NONE
 from prettytable import PrettyTable
 
+
+if sys.version_info >= (3, 11):
+    import tomllib
+else:
+    import tomli as tomllib
```
reactive-firewall: Looks lie this causes a minor coverage regression due to [changes around line 56](https://app.codecov.io/gh/raimon49/pip-licenses/pull/214/blob/piplicenses.py?dropdown=coverage#L56) which is now only covered by python versions pre-3.11.

This may actually be an issue with how coverage tracks these lines as coverage is reported for python 3.9 and 3.10 to codecov.


Suggested fix is to update coverage config to ignore `tomli` lines as it is known to be flaky

piplicenses.py
```diff
@@ -178,15 +183,16 @@ def normalize_pkg_name(pkg_name: str) -> str:
...
     "pip",
     "prettytable",
     "wcwidth",
     "setuptools",
-    "tomli",
     "wheel",
-)
+]
+if sys.version_info < (3, 11):
+    SYSTEM_PACKAGES.append("tomli")
```
reactive-firewall: Same issue as around line 56 but fix here is to just mark the whole conditional as no-coverage.

piplicenses.py
cdce8p: Yeah, the coverage is only evaluated for Python 3.11. Added `# pragma: no cover` for the two blocks.

stefan6419846: To be honest, adding the pragma marker might work, but just collecting coverage data for each Python version and merging them together for coverage analysis usually is the better approach.

cdce8p: > To be honest, adding the pragma marker might work, but just collecting coverage data for each Python version and merging them together for coverage analysis usually is the better approach.

👍🏻 I agree. Guess that's a task for a future PR.

reactive-firewall: > > To be honest, adding the pragma marker might work, but just collecting coverage data for each Python version and merging them together for coverage analysis usually is the better approach.
> 
> 👍🏻 I agree. Guess that's a task for a future PR.

:+1: I second the idea; and agree that can be a separate PR as it may involve verifying the coverage module can track the import alias (I vaugly recall there being an edge-case about some aliases, so I want to find an ref the docs to be sure before resolving this point)

CI/coverage details aside, rest of the changes look good to me.

I'll make sure there is a GHI for this task, (and might set it up as a "good-first-issue" if I'm over-thinking it) after I have double-checked the docs.

PR #216, opened 2024-10-12
Fixed: KeyError caused due to the same element getting removed from a set twice

anmol-aidora:
```
for item_a in set_a:
    for item_b in set_b:
        if <some condition:True>:
            set_a.remove(item_a)
```
As you can see, if there are multiple items in set_b for which the condition evaluates to True, set_a.remove(item_a) will run twice, which will cause KeyError for second time onwards.

To solve this problem, I am checking whether the item_a is there in the set_a before removing. If it is, then remove, if it isn't then continue without removing as we have already removed it

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -435,6 +435,8 @@ def case_insensitive_partial_match_set_diff(set_a, set_b):
     for item_a in set_a:
         for item_b in set_b:
             if item_b.lower() in item_a.lower():
+                if item_a not in uncommon_items:
+                    continue
                 uncommon_items.remove(item_a)
     return uncommon_items
 


stefan6419846:
There should probably be a test as well which covers this case. Additionally, using `uncommon_items.discard(item_a)` directly should do the same with less code.

anmol-aidora:
> There should probably be a test as well which covers this case. Additionally, using `uncommon_items.discard(item_a)` directly should do the same with less code.

Thanks @stefan6419846 for pointing out this method. For the uninitiated:
The `discard()` method removes the specified item from the set. This method is different from the `remove()` method, because the `remove()` method will raise an error if the specified item does not exist, and the `discard()` method will not.

I am not sure how to write a test case for this. I will take that up later

piplicenses.py
```diff
@@ -435,7 +435,7 @@ def case_insensitive_partial_match_set_diff(set_a, set_b):
     for item_a in set_a:
         for item_b in set_b:
             if item_b.lower() in item_a.lower():
-                uncommon_items.remove(item_a)
+                uncommon_items.discard(item_a)
     return uncommon_items
```
reactive-firewall, line 439: :broom: NIT

<details><summary>Consider collecting then discarding, don't modify while iterating</summary>

**Partial fix**

https://github.com/raimon49/pip-licenses/blob/3fbf20f9d757d2b9298dfc97af1bd3a987124c98/piplicenses.py#L433-L440

I agree with the use of `discard` in this case (was new to me too). However I think the unique usage of the `copy()` before then modifying is strange when the rest of the code uses a fresh new `set()`.

**Potential Issues Still Present**
While using discard can prevent `KeyError`, the original logic still has a potential issue related to modifying the set while iterating over it. If `itemₐ` is removed during the iteration, it can lead to unexpected behavior, such as skipping some items or not processing them correctly.

**suggested approach**

```python
    """
    Returns a set of items from set_a that do not have any case-insensitive partial matches in set_b.

    Args:
        set_a (set): A set of strings to compare.
        set_b (set): A set of strings to check for partial matches.

    Returns:
        set: A set of items from set_a that do not match any items in set_b.

    Example:
        >>> case_insensitive_partial_match_set_diff({'BSD License', 'MIT License'}, {'BSD'})
        {'MIT License'}
        
        >>> case_insensitive_partial_match_set_diff({'BSD', 'BSD License'}, {'BSD'})
        set()

        >>> case_insensitive_partial_match_set_diff({'Hello', 'World'}, {'hello'})
        {'World'}

        >>> case_insensitive_partial_match_set_diff({'Duplicate', 'duplicate', 'Unique'}, {'unique'})
        {'Duplicate', 'duplicate'}

        >>> case_insensitive_partial_match_set_diff({'Test', 'Example'}, {'Sample', 'Test'})
        {'Example'}

        >>> case_insensitive_partial_match_set_diff(set(), {'Empty'})
        set()

        >>> case_insensitive_partial_match_set_diff({'A', 'B', 'C'}, {'D', 'E'})
        {'A', 'B', 'C'}
    """
    uncommon_items = set_a.copy()
    to_remove = set()  # Create a set to track items to remove

    for item_a in set_a:
        for item_b in set_b:
            if item_b.lower() in item_a.lower():
                to_remove.add(item_a)  # Add to the removal set

    uncommon_items.difference_update(to_remove)  # Remove all at once
    return uncommon_items
```

</details>

Consider the following:

```suggestion
    to_remove = set()  # Create a set to track items to remove
    for item_a in set_a:
        for item_b in set_b:
            if item_b.lower() in item_a.lower():
                to_remove.add(item_a)  # Add to the removal set
    uncommon_items.difference_update(to_remove)  # Remove all from set at once
    return uncommon_items
```

piplicenses.py
reactive-firewall, line 1: Regarding test-code; here are some hints:

as a docstring doc-test:
```python
    """
    Returns a set of items from set_a that do not have any case-insensitive partial matches in set_b.

    Args:
        set_a (set): A set of strings to compare.
        set_b (set): A set of strings to check for partial matches.

    Returns:
        set: A set of items from set_a that do not match any items in set_b.

    Example:
        >>> case_insensitive_partial_match_set_diff({'BSD License', 'MIT License'}, {'BSD'})
        {'MIT License'}
        
        >>> case_insensitive_partial_match_set_diff({'BSD', 'BSD License'}, {'BSD'})
        set()

        >>> case_insensitive_partial_match_set_diff({'Hello', 'World'}, {'hello'})
        {'World'}

        >>> case_insensitive_partial_match_set_diff({'Duplicate', 'duplicate', 'Unique'}, {'unique'})
        {'Duplicate', 'duplicate'}

        >>> case_insensitive_partial_match_set_diff({'Test', 'Example'}, {'Sample', 'Test'})
        {'Example'}

        >>> case_insensitive_partial_match_set_diff(set(), {'Empty'})
        set()

        >>> case_insensitive_partial_match_set_diff({'A', 'B', 'C'}, {'D', 'E'})
        {'A', 'B', 'C'}
    """
```

reactive-firewall (changes requested):
> [!IMPORTANT]
> **Disclaimer:** I am just a fellow user of `pip-licenses` and have no official association with the project or its maintainers. Any feedback or comments I provide are based on my personal experience and are offered AS-IS. I cannot guarantee their accuracy or effectiveness, and they should not be considered official support or advice. Please reach out to the maintainers for authoritative guidance.

:+1: On the right track; Here is some review hints for the test-code and nit-picky improvements in hopes they help :bow:

reactive-firewall:
Closing as Superseded by PR #238

PR #217, opened 2024-10-12
Update pinned dev requirements

cdce8p:


piplicenses.py
```diff
@@ -440,9 +442,11 @@ def case_insensitive_partial_match_set_intersect(set_a, set_b):
 def case_insensitive_partial_match_set_diff(set_a, set_b):
     uncommon_items = set_a.copy()
     for item_a in set_a:
+        item_a_lower = item_a.lower()
         for item_b in set_b:
-            if item_b.lower() in item_a.lower():
+            if item_b.lower() in item_a_lower:
                 uncommon_items.remove(item_a)
+                break
     return uncommon_items
```
cdce8p, line 450: Updating `cryptography` uncovered a `KeyError` here. With PEP 639 they updated their license string to `Apache-2.0 OR BSD-3-Clause`. The test case uses `--partial-match --allow-only=...` with `Bsd` and `Apache` thus matching the license string twice. Without the added `break`, the code would try to remove the licesne from `uncommon_items` twice resulting in a KeyError.

https://github.com/raimon49/pip-licenses/blob/3fbf20f9d757d2b9298dfc97af1bd3a987124c98/test_piplicenses.py#L893-L906

cdce8p:
Changed the target branch to `dev-5.5.0`, update the pinned dependencies again and fixed the (now) failing test cases.

piplicenses.py
cdce8p: Seems I also needed to fix #216 / #224 😅
Let me know if I should revert the change here or how you'd like to proceed.

reactive-firewall:
Closing as Included-in and Superseded-by PR #238

PR #219, opened 2024-10-12
Update build-system

cdce8p:
Since setuptools `70.1` the `bdist_wheel` command is shipped with setuptools directly. It's no longer necessary to specify `wheel` as a build system requirement.


cdce8p:
Changed the target branch to `dev-5.5.0`.

PR #220, opened 2024-10-12
Remove prettytable stubs

cdce8p:
`prettytable` includes typing information and a `py.typed` file since `3.5.0`. The stubs are no longer necessary.

https://github.com/jazzband/prettytable/releases/tag/3.5.0


cdce8p:
The stubs are quite incomplete. Let's leave it as is for the time being.

PR #221, opened 2024-10-14
Add support for multi-line allow-only and fail-on arguments

anmol-aidora:
When setting the `allow-only` and `fail-on` arguments, the list can get very long very fast, especially when switching `partial` to `false`. There is a need to make sure multi-line arguments work properly.

Currently when passing multi-line allow-only and fail-on arguments in `pyproject.toml`, the `fail_on_licenses` and `allow_only_licenses` sets were found to contain the empty string `''`.  

e.g. 
```
allow-only="""
MIT;Apache;
"""
```
sets allow_only_licenses to:
```
{'', 'Apache', 'MIT'}
```

The code checks for the same and removes it from the sets, such that there is no empty string:
```
{'Apache', 'MIT'}
```

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -284,11 +284,11 @@ def get_packages(
 
     fail_on_licenses = set()
     if args.fail_on:
-        fail_on_licenses = set(map(str.strip, args.fail_on.split(";")))
+        fail_on_licenses = {license.strip() for license in args.fail_on.split(";") if license.strip()}
 
     allow_only_licenses = set()
     if args.allow_only:
-        allow_only_licenses = set(map(str.strip, args.allow_only.split(";")))
+        allow_only_licenses = {license.strip() for license in args.allow_only.split(";") if license.strip()}
 
     for pkg in pkgs:
         pkg_name = normalize_pkg_name(pkg.metadata["name"])


piplicenses.py
```diff
@@ -284,11 +284,11 @@ def get_python_sys_path(executable: str) -> list[str]:
 
     fail_on_licenses = set()
     if args.fail_on:
-        fail_on_licenses = set(map(str.strip, args.fail_on.split(";")))
+        fail_on_licenses = {license.strip() for license in args.fail_on.split(";") if license.strip()}
```
reactive-firewall, line 287:
```suggestion
        fail_on_licenses = set(filter(None, map(str.strip, args.fail_on.split(";"))))
```

piplicenses.py
```diff
@@ -284,11 +284,11 @@ def get_python_sys_path(executable: str) -> list[str]:
...
 
     allow_only_licenses = set()
     if args.allow_only:
-        allow_only_licenses = set(map(str.strip, args.allow_only.split(";")))
+        allow_only_licenses = {license.strip() for license in args.allow_only.split(";") if license.strip()}
```
reactive-firewall, line 291:
```suggestion
        allow_only_licenses = set(filter(None, map(str.strip, args.allow_only.split(";"))))
```

reactive-firewall:
Superseded by #287

PR #237, opened 2025-10-12
[ci] Upload coverage for all python versions

### LINKED ISSUE #236: Chore - CI/CD modernization and cross-python coverage submissions
This is a chore related to PR #214 where coverage of the `tomli` imports regressed slightly. It was suggested that we look-into the coverage submissions as they may not be combining cross-python results at the moment.

- [x] Ref Docs:
   * [tomli compatibility](https://github.com/hukkin/tomli/blob/master/README.md#building-a-tomlitomllib-compatibility-layer) (provided by PR #214)
   * [tomllib](https://docs.python.org/3/library/tomllib.html) (e.g., docs show added python3.11)
   * [codecov action](https://github.com/codecov/codecov-action/blob/main/README.md#arguments)
   * [Docs confirming obscure import issues were fixed in coverage](https://coverage.readthedocs.io/en/7.10.7/changes.html#version-7-6-10-2024-12-26) (so should not need to worry about the import alias stuff :tada: sorry for any confusion)
   * [Branch coverage edge-case docs](https://coverage.readthedocs.io/en/7.10.7/branch.html#structurally-partial-branches) (for completeness only)

- [x] CI implementation changes
  - [x] remove pragma workaround - provided by #PR #237
  - [ ] Implement cross-python coverage reporting

- [ ] Document rationale of solution (e.g., a short "here's why we did xyz" for posterity and completeness)

CC: @stefan6419846
CC: @cdce8p

cdce8p:
Fixes #236

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -37,11 +37,24 @@ jobs:
         run: |
           mkdir -p .mypy_cache
           mypy --install-types --non-interactive .
+      - name: Update coverage artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-${{ matrix.python-version }}
+          path: coverage.xml
+
+  coverage:
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/checkout@v4
+      - name: Download all coverage artifacts
+        uses: actions/download-artifact@v5
+        with:
+          pattern: coverage-*
       - name: Upload code coverage to Codecov
-        if: ${{ matrix.python-version==3.11 }}
         uses: codecov/codecov-action@v4
         with:
-          files: ./coverage.xml
           name: codecov-umbrella
           fail_ci_if_error: true
           verbose: true
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -52,7 +52,7 @@ from prettytable import PrettyTable
 
 if sys.version_info >= (3, 11):
     import tomllib
-else:  # pragma: no cover
+else:
     import tomli as tomllib
 
 if TYPE_CHECKING:  # pragma: no cover
@@ -191,7 +191,7 @@ SYSTEM_PACKAGES = [
     "setuptools",
     "wheel",
 ]
-if sys.version_info < (3, 11):  # pragma: no cover
+if sys.version_info < (3, 11):
     SYSTEM_PACKAGES.append("tomli")
 
 LICENSE_UNKNOWN = "UNKNOWN"


.github/workflows/python-package.yml
```diff
@@ -37,11 +37,24 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-${{ matrix.python-version }}
+          path: coverage.xml
+
+  coverage:
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
+      - uses: actions/checkout@v4
```
reactive-firewall, line 50: :thinking: We should use the latest checkout action while we're at it.

Consider with sha pinning:

```suggestion
      - name: Checkout code
        uses: actions/checkout@08c6903cd8c0fde910a37f88322edcfb5dd907a8 # v5.0.0
        with:
          persist-credentials: false
```

or at-least refactor to latest:

```suggestion
      - uses: actions/checkout@v5
```

I'll leave the style up-to you.

.github/workflows/python-package.yml
```diff
@@ -37,11 +37,24 @@ jobs:
         run: |
           mkdir -p .mypy_cache
           mypy --install-types --non-interactive .
+      - name: Update coverage artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-${{ matrix.python-version }}
+          path: coverage.xml
```
reactive-firewall, line 44: :broom: NIT - also upload failed tests to ensure comparisons are possible for fixes

**Rationale**:
The default if condition is success based, however even failed tests can upload coverage reports. Coverage reports must be available on base commits of PRs for Codecov to compare patch coverage, so we should also upload the failed test coverage when possible. For now prefer `if: ${{ !cancelled() }}`

also while here consider this more robust implementation:

```suggestion
      - name: Update coverage artifact
        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
        if: ${{ !cancelled() }}  # delay the outcome
        with:
          name: coverage-${{ matrix.python-version }}
          path: coverage.xml
          overwrite: true  # don't error on retries
          if-no-files-found: error  # handle delayed outcome here
          retention-days: 1  # only used for CI/CD
```

Caveat: Current implementation is already acceptable, so my NIT suggestion is stylistic and not strictly necessary for this PR.

piplicenses.py
reactive-firewall, line 1: ## :+1: LGTM:

Specifically the changes to `piplicenses.py` are approved as-is and indeed address #236

.github/workflows/python-package.yml
```diff
@@ -37,11 +37,24 @@ jobs:
...
+      - name: Download all coverage artifacts
+        uses: actions/download-artifact@v5
+        with:
+          pattern: coverage-*
       - name: Upload code coverage to Codecov
-        if: ${{ matrix.python-version==3.11 }}
         uses: codecov/codecov-action@v4
         with:
-          files: ./coverage.xml
           name: codecov-umbrella
```
reactive-firewall, line 58: Changes to codecov action could be further improved.

While the removed condition for python v3.11 does seem to address #236, I'm not convinced the different coverage reports are indeed each being uploaded, and could prove a hinderance to maintaining the coverage data properly.

I'd prefer a solution that either merges the coverage files after downloading the artifacts (albeit complex, it allows the re-use of the `codecov-umprella` grouping) or explicit flagging of each python-version.

If you choose to merge the coverage reports use the merge option in the download-artifacts action and then use python coverage to combine and generate a new xml.

... or we can merge the upload into the matrix job:

Consider updating the version and options with something like:

```suggestion
      if: ${{ !cancelled() && (github.repository == 'raimon49/pip-licenses') }}
      uses: codecov/codecov-action@5a1091511ad55cbe89839c7260b706298ca349f7 # v5.5.1
      with:
        token: ${{ secrets.CODECOV_TOKEN }}  # more stable than the env pass IMHO
        job_code: ${{ github.sha || 'PR' }}-${{ github.run_id }}-${{ github.run_number || '0' }}  # group all together
        # override_commit: ${{ github.sha }}
        report_type: 'coverage'
        slug: 'raimon49/pip-licenses'
        files: ./coverage.xml,./coverage-*/*.xml  # should still restrict this somewhat
        directory: ${{ github.workspace }}
        # flags: ${{ matrix.python-version || 'unknown' }} # could allow per-python groups
        name: codecov-umbrella-${{ matrix.python-version }}
```

This nuance is part of why I felt GHI #236 warranted its own ticket. I'll leave the direction up to you (e.g., revert and flag by version, or continue but handle merging).

.github/workflows/python-package.yml
cdce8p: Not sure I'd change anything here.
- SHA pinning: Yes that's valid but might be best to do separately as all other actions aren't pinned as well.
- `!cancelled()`: I don't think that makes sense. If there are test errors, those need to be resolved first anyway. `main` should ideally always pass.
- `overwrite`, `if-no-files-found` and `retention-days` yes they could be added but are really just minor.

.github/workflows/python-package.yml
cdce8p: As mentioned above, I think it would be best if that's done in a separate PR later. The codecov-action is also up-of-date.

With this one I strictly focus on adding the coverage for all python versions.

.github/workflows/python-package.yml
cdce8p: Tbh I don't think it's worth the effort. You can checks the logs from that last run to see all are being uploaded.
https://github.com/raimon49/pip-licenses/actions/runs/18443482715/job/52547008871#step:4:227

If at some point that stops working, the coverage result would change instantly. Though I don't think it will given that quite a lot of projects are relying on that exact behavior.

cdce8p:
@reactive-firewall I left comments on each of your review points. All in all I don't think there is a need to overcomplicate things for a project of this size. The SHAs for each action and updating the actions might be better placed in a separate PR.

_I removed my assignment as I'm the author of the PR anyway._

.github/workflows/python-package.yml
reactive-firewall: Thanks for calling out the logs. I see it is already finding the coverage just fine, and don't see this a blocker for this PR anymore.

TL;DR - However, as I do see it as worth the effort I'll probably implement the flag approach at some point to provide visibility between python version coverage.

PR #234, opened 2025-08-03
Add support for PEP 639 change from 'License' to 'License-Expression' in importlib.metadata

bskinn:
[PEP 639] made some changes to the packaging metadata standard, including deprecating [`License`] in favor of [`License-Expression`]. This has led to various packages' license failing to locate for releases after their build system rolled over onto PEP 639-compliant tooling.

This PR adds a second `lambda` to the lookup for the `"license"` metadata key, pulling the `"license-expression"` metadata entry, so that metadata with either key/field can be properly processed.

Given that `License-Expression` is now the standard and `License` is deprecated, there might be semantic reasons to put the check for `"license-expression"` first; I figure that's your call.

Example environment:

```bash
$ pip list
Package            Version Editable project location
------------------ ------- ----------------------------
black              25.1.0
build              1.3.0
click              8.2.1
coverage           7.10.1
docutils           0.22
iniconfig          2.1.0
isort              6.0.1
mypy_extensions    1.1.0
packaging          25.0
pathspec           0.12.1
pip                25.2
pip-licenses       5.0.0   .../pip-licenses
pip-tools          7.5.0
platformdirs       4.3.8
pluggy             1.6.0
prettytable        3.16.0
pycodestyle        2.14.0
Pygments           2.19.2
pyproject_hooks    1.2.0
pytest             8.4.1
pytest-cov         6.2.1
pytest-pycodestyle 2.5.0
setuptools         80.9.0
tomli              2.2.1
tomli_w            1.2.0
wcwidth            0.2.13
wheel              0.45.1
```

Before the fix (my arrows added):

```bash
$ pip-licenses
 Name                Version  License
 Pygments            2.19.2   BSD License
 black               25.1.0   MIT License
 build               1.3.0    UNKNOWN          <<<<<
 click               8.2.1    UNKNOWN          <<<<<
 coverage            7.10.1   Apache-2.0
 docutils            0.22     BSD License; GNU General Public License (GPL); Public Domain
 iniconfig           2.1.0    MIT License
 isort               6.0.1    MIT License
 mypy_extensions     1.1.0    UNKNOWN          <<<<<
 packaging           25.0     Apache Software License; BSD License
 pathspec            0.12.1   Mozilla Public License 2.0 (MPL 2.0)
 pip-tools           7.5.0    BSD License
 platformdirs        4.3.8    MIT License
 pluggy              1.6.0    MIT License
 pycodestyle         2.14.0   MIT
 pyproject_hooks     1.2.0    MIT License
[...]

PR #224, opened 2024-12-16
Fix KeyError on double deletion

flontis:

PR #238, opened 2025-10-13
Refactor Partial Matching

### LINKED ISSUE #231: Partial matching goes wrong when partial match
# Minimum working example
(assuming [uv](https://docs.astral.sh/uv/getting-started/installation/) is installed)
```shell
# setup the environment
mkdir sandbox
cd sandbox
uv init
uv add pillow==10.4.0 Cerberus pip-licenses
```

Now run `pip-licenses` and allow all licenses with HPND and ISC.
```shell
pip-licenses --partial-match --allow-only="HPND;ISC"
```

I get the following error:

```shell
Traceback (most recent call last):
  File "/mnt/c/Users/boerp/D-schijf/Projects/sandbox/.venv/bin/pip-licenses", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/mnt/c/Users/boerp/D-schijf/Projects/sandbox/.venv/lib/python3.12/site-packages/piplicenses.py", line 1157, in main
    output_string = create_output_string(args)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/Users/boerp/D-schijf/Projects/sandbox/.venv/lib/python3.12/site-packages/piplicenses.py", line 693, in create_output_string
    table = create_licenses_table(args, output_fields)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/Users/boerp/D-schijf/Projects/sandbox/.venv/lib/python3.12/site-packages/piplicenses.py", line 369, in create_licenses_table
    for pkg in get_packages(args):
  File "/mnt/c/Users/boerp/D-schijf/Projects/sandbox/.venv/lib/python3.12/site-packages/piplicenses.py", line 345, in get_packages
    uncommon_licenses = case_insensitive_partial_match_set_diff(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/mnt/c/Users/boerp/D-schijf/Projects/sandbox/.venv/lib/python3.12/site-packages/piplicenses.py", line 438, in case_insensitive_partial_match_set_diff
    uncommon_items.remove(item_a)
KeyError: 'Historical Permission Notice and Disclaimer (HPND)'
```

# Expected result:
I expected no problems, since both licenses are allowed. In any case the program should not crash but should return a decent message.

# Problem analysis
The license "Historical Permission Notice and Disclaimer (HPND)" contains the letters "isc". In function `case_insensitive_partial_match_set_diff` the variable `uncommon_items` is already emptied by the first hit. When the second hit occurs, there is nothing to remove anymore (line 441). 

Image

# Proposed solution
Add a check for an empty `uncommon_items` set:
```python
    uncommon_items = set_a.copy()
    for item_a in set_a:
        for item_b in set_b:
            if item_b.lower() in item_a.lower():
                uncommon_items.remove(item_a)
            if len(uncommon_items) == 0:
                break
    return uncommon_items
```
# Used version
I was able to reproduce this problem with v4.5.1 and v5.0.0.

reactive-firewall:
# Patch Notes

## Impacted GHI

 - [x] closes #231

## Impacted PRs

 * includes and supersedes #216
 * includes and supersedes #217
 * includes and supersedes #224
 * supersedes #234

---

> [!IMPORTANT]
> :bow: Big thanks to all of the efforts from multiple contributors on this patch!

### Merge Notes

<details><summary>Full Merge Details</summary>

By @anmol-aidora
  * patch-2:
    ~Replaced element check logic with built in set.discard method~ _(Resolved)_
    Fixed: KeyError caused due to the same element getting removed from a set twice

By @anmol-aidora
  * fix-keyerror:
    ~Fix KeyError on double deletion~ _(Resolved)_

By @cdce8p 
  * update-requirements:
    Update pinned requirements again
    Add changelog entry
    Fix KeyError if license matches multiple allow-only licenses
    Fix test case
    Update pinned dev requirements

By @reactive-firewall 
  * Closes GHI #231 (and related issues with modifing sets within loops)
  * Improved related testing slightly with extra test-cases

  * **Resolves Conflicts** in:
    `piplicenses.py`

</details>

### DIFF
--- a/test_piplicenses.py
+++ b/test_piplicenses.py
@@ -1280,6 +1280,10 @@ def test_case_insensitive_partial_match_set_diff():
     result = case_insensitive_partial_match_set_diff(set_a, set_b)
     assert result == set(), "The function should match partials (inclusively)."
 
+    set_a = {"Apache-2.0 OR BSD-3-Clause"}
+    set_b = {"BSD", "Apache"}
+    result = case_insensitive_partial_match_set_diff(set_a, set_b)
+    assert result == set(), "Multiple matches shouldn't crash."
     set_a = {"Duplicate", "duplicate", "Unique"}
     set_b = {"unique"}
     result = sorted(case_insensitive_partial_match_set_diff(set_a, set_b))


cdce8p:
I haven't looked at the changes yet but is the target branch correct? Shouldn't it be dev-5.5.0?

reactive-firewall:
> I haven't looked at the changes yet but is the target branch correct? Shouldn't it be dev-5.5.0?

Yup, and fixed now.

piplicenses.py
```diff
@@ -440,11 +440,62 @@ def case_insensitive_partial_match_set_intersect(set_a, set_b):
...
+
+        >>> case_insensitive_partial_match_set_diff(set(), {'Empty'})
+        set()
+
+        >>> sorted(case_insensitive_partial_match_set_diff(
+        ...     {'A', 'B', 'C'}, {'D', 'E'}
+        ... ))
+        ['A', 'B', 'C']
+
+    """
```
cdce8p, line 446: Not sure the documentation really needs to be this extensive. It's only an internal helper function with less than 10 LOC. In my experience if the docstring is 5 times as long as the actual code, I'm probably doing something wrong.

piplicenses.py
```diff
@@ -440,11 +440,62 @@ def case_insensitive_partial_match_set_intersect(set_a, set_b):
...
     uncommon_items = set_a.copy()
+    to_remove = set()  # Create a set to track items to remove
     for item_a in set_a:
+        item_a_lower = item_a.lower()  # change case once & use for sub loop
         for item_b in set_b:
-            if item_b.lower() in item_a.lower():
-                uncommon_items.remove(item_a)
+            if item_b.lower() in item_a_lower:
+                to_remove.add(item_a)  # Add to the removal set
+    uncommon_items.difference_update(to_remove)  # Remove all at once
```
cdce8p: Why not use `discard` as suggested in #216 or even my check with `remove(...)` and `break`. Both work and are easier to follow than creating a separate set and using `difference_update`.

Besides that, I don't think it's necessary to add comments to every line. In an ideal world code should be self documenting. Any reasonable dev should be able to understand what's happening just by looking at 10 lines. If not, you can always ask an LLM for an explanation.

piplicenses.py
reactive-firewall: > Not sure the documentation really needs to be this extensive. It's only an internal helper function with less than 10 LOC. 

While I'll agree with the well known quote:

> Simplicity does not precede complexity, but follows it.
- [Alan Perlis](https://www.azquotes.com/author/11531-Alan_Perlis)

I appreciate the feedback. Could you please clarify which specific changes you recommend? I'm happy to compare alternatives if you can point me to the suggested approach.

> In my experience if the docstring is 5 times as long as the actual code, I'm probably doing something wrong

TL;DR — My experience differs. If something doesn't require documentation, it may be incomplete or not useful. I prefer TDD: if I can't describe the test case, I can't be confident it works deterministically.

piplicenses.py
reactive-firewall: > Besides that, I don't think it's necessary to add comments to every line. In an ideal world code should be self documenting.

And because I didn't change code for the ideal case, (this change is to fix a bug) I'll quote an expert:

> Good code is its own best documentation. As you're about to add a comment, ask yourself, "How can I improve the code so that this comment isn't needed?" Improve the code and then document it to make it even clearer.
- [Steve McConnell](https://www.inspiringquotes.us/author/2732-steve-mcconnell)

I won't remove the docstring simply because it explains something we already know; it was written to specify the required behavior, not to restate obvious facts.

Regarding the inline comments: what specifically do you consider better? If you can show the alternative, I'm happy to evaluate it.

> Any reasonable dev should be able to understand what's happening just by looking at 10 lines. If not, you can always ask an LLM for an explanation.

I agree in principle (even if the LLM hallucinates). To clarify the risk here, I used an LLM to summarize the core issue:

> Short answer: it's a bad idea to modify a Python set (or other container) while iterating because it either raises a RuntimeError or produces incorrect/undefined iteration behavior.
> 
> Why:
> - **Sets and dicts:** Python detects size changes and raises `RuntimeError: "set changed size during iteration"` (or `"dictionary changed size during iteration"`) because mutating the container invalidates the iterator.
> - **Lists:** Python usually doesn't raise, but modifying a list while iterating will change indexes and can skip or repeat elements, producing incorrect results.

As noted in PR #216 [in this comment (expand the whole comment if confused)](https://github.com/raimon49/pip-licenses/pull/216#discussion_r2188841848):

> **Potential Issues Still Present**
> While using discard can prevent KeyError, the original logic still has a potential issue related to modifying the set while iterating over it. If item_a is removed during the iteration, it can lead to unexpected behavior, such as skipping some items or not processing them correctly.

Those behaviors tend to be flaky and hard to test reliably, which is a maintenance concern. For that reason I avoided approaches that modify the set during iteration: why accept "usually" correct when we can make it deterministic? I remain unconvinced that mutating during iteration is the safer choice.

piplicenses.py
cdce8p: > Could you please clarify which specific changes you recommend? I'm happy to compare alternatives if you can point me to the suggested approach.

```suggestion
    """
    Return items from set_a without case-insensitive partial matches from items in set_b.
    """
```

At most I'd include the first example as well
```py
    Example:
        >>> case_insensitive_partial_match_set_diff(
        ...     {'BSD License', 'MIT License'}, {'BSD'}
        ... )
        {'MIT License'}
```

Everything else is either a duplication, unnecessary or handled in the test cases.

> > TL;DR — My experience differs. If something doesn't require documentation, it may be incomplete or not useful. I prefer TDD: if I can't describe the test case, I can't be confident it works deterministically.

That's fine but isn't that what test cases are for. This is still a docstring after all.

piplicenses.py
cdce8p: > > **Potential Issues Still Present**
> > While using discard can prevent KeyError, the original logic still has a potential issue related to modifying the set while iterating over it. If item_a is removed during the iteration, it can lead to unexpected behavior, such as skipping some items or not processing them correctly.
> 
> Those behaviors tend to be flaky and hard to test reliably, which is a maintenance concern. For that reason I avoided approaches that modify the set during iteration: why accept "usually" correct when we can make it deterministic? I remain unconvinced that mutating during iteration is the safer choice.

The items aren't removed / discarded from `set_a` though. `uncommon_items` is a `copy` of `set_a`.

> Regarding the inline comments: what specifically do you consider better? If you can show the alternative, I'm happy to evaluate it.

```suggestion
    uncommon_items = set_a.copy()
    for item_a in set_a:
        item_a_lower = item_a.lower()
        for item_b in set_b:
            if item_b.lower() in item_a_lower:
                uncommon_items.remove(item_a)
                break
    return uncommon_items
```

piplicenses.py
reactive-firewall: > That's fine but isn't that what test cases are for. This is still a docstring after all.

:see_no_evil: docstrings can be both documentation and [doctests](https://docs.python.org/3/library/doctest.html); 

> Everything else is either a duplication, unnecessary or handled in the test cases.

:thinking: Good point! This project's convention is to use the separate tests cases. We'll go with the one-line description you suggested, on the rationale it seems more consistent with the project's existing docstrings.

piplicenses.py
reactive-firewall: > The items aren't removed / discarded from `set_a` though. `uncommon_items` is a `copy` of `set_a`.

:facepalm: hu? Not sure how I missed that! But after double checking I concur you are right, main difference is your suggestion with the break is faster for small sets, and mine is faster for huge sets. That makes yours clearly more suited to licences (where sets are typically smaller than 1k items)

We'll also go with your suggestion here. :bow: Thanks again for your feedback and for catching this.

piplicenses.py
```diff
@@ -440,11 +440,62 @@ def case_insensitive_partial_match_set_intersect(set_a, set_b):
...
+        set()
+
+        >>> sorted(case_insensitive_partial_match_set_diff(
+        ...     {'A', 'B', 'C'}, {'D', 'E'}
+        ... ))
+        ['A', 'B', 'C']
+
+    """
     uncommon_items = set_a.copy()
+    to_remove = set()  # Create a set to track items to remove
```
reactive-firewall:
```suggestion
```

piplicenses.py
```diff
@@ -440,11 +440,62 @@ def case_insensitive_partial_match_set_intersect(set_a, set_b):
...
     for item_a in set_a:
+        item_a_lower = item_a.lower()  # change case once & use for sub loop
         for item_b in set_b:
-            if item_b.lower() in item_a.lower():
-                uncommon_items.remove(item_a)
+            if item_b.lower() in item_a_lower:
+                to_remove.add(item_a)  # Add to the removal set
+    uncommon_items.difference_update(to_remove)  # Remove all at once
     return uncommon_items
```
reactive-firewall, line 454:
```suggestion
            if item_b.lower() in item_a_lower:
                uncommon_items.discard(item_a)
                break
    return uncommon_items
```

piplicenses.py
```diff
@@ -440,11 +440,16 @@ def case_insensitive_partial_match_set_intersect(set_a, set_b):
 
 
 def case_insensitive_partial_match_set_diff(set_a, set_b):
+    """
+    Return items from set_a without case-insensitive partial matches from items in set_b.
```
reactive-firewall: Wrap lines for linter checks.

```suggestion
    Return items from set_a without case-insensitive partial matches
    from items in set_b.
```

test_piplicenses.py
```diff
@@ -1216,3 +1216,89 @@ def test_pyproject_toml_args_parsed_correctly():
...
+    result = case_insensitive_partial_match_set_diff(set_a, set_b)
+    assert result == {
+        "MIT License"
+    }, "The function should match partials (exclusively)."
+
+    set_a = {"BSD", "BSD License"}
+    set_b = {"BSD"}
+    result = case_insensitive_partial_match_set_diff(set_a, set_b)
+    assert result == set(), "The function should match partials (inclusively)."
+
```
cdce8p, line 1282:
```suggestion

    set_a = {"Apache-2.0 OR BSD-3-Clause"}
    set_b = {"BSD", "Apache"}
    result = case_insensitive_partial_match_set_diff(set_a, set_b)
    assert result == set(), "Multiple matches shouldn't crash."
```
A test case which would fail without the change would be nice.

cdce8p (approved):
Looks good, one last suggestions. Feel free to merge afterwards.

PR #261, opened 2026-01-17
Apply Ruff rules

DimitriPapadopoulos:
Rules are not enforced in CI for now.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -807,21 +807,17 @@ def create_warn_string(args: CustomNamespace) -> str:
 
     if args.with_license_file and args.format_ != FormatArg.JSON:
         message = warn(
-            (
-                "Due to the length of these fields, this option is "
-                "best paired with --format=json."
-            )
+            "Due to the length of these fields, this option is "
+            "best paired with --format=json."
         )
         warn_messages.append(message)
 
     if args.summary and (args.with_authors or args.with_urls):
         message = warn(
-            (
-                "When using this option, only --order=count or "
-                "--order=license has an effect for the --order "
-                "option. And using --with-authors and --with-urls "
-                "will be ignored."
-            )
+            "When using this option, only --order=count or "
+            "--order=license has an effect for the --order "
+            "option. And using --with-authors and --with-urls "
+            "will be ignored."
         )
         warn_messages.append(message)
 


test_piplicenses.py
```diff
@@ -1189,7 +1182,7 @@ def test_extract_homepage_home_page_set() -> None:
     metadata = MagicMock()
     metadata.get.return_value = "Foobar"
 
-    assert "Foobar" == extract_homepage(metadata=metadata)  # type: ignore
+    assert "Foobar" == extract_homepage(metadata=metadata)
```
stefan6419846, line 1185: How is this change related to *ruff*? Isn'T this about type checking? If I remember correctly, the ignore had been required as a `MagicMock` is no "real" package metadata when typed properly.

DimitriPapadopoulos: This Ruff rule requests "_specific rule codes when ignoring type issues_". I chose to remove `# type: ignore` altogether because mypy didn't raise any error without it on my machine. But then I'm probably using the wrong version of mypy. Let me have a look.

DimitriPapadopoulos: You're right, I had to change another `# type: ignore` to `# type: ignore[arg-type]` to avoid this MyPy error:
```
error: Argument 1 has incompatible type "PackageMetadata"; expected "Message[str, str]"  [arg-type]
```

But this one is not needed as far as I can see.

piplicenses.py
```diff
@@ -630,12 +624,12 @@ def esc_quotes(val: bytes | str) -> str:
 
         lines: list[str] = []
         formatted_header = ",".join(
-            ['"%s"' % (esc_quotes(val),) for val in self._field_names]
+            [f'"{esc_quotes(val)}"' for val in self._field_names]
```
reactive-firewall, line 627: **Additional Comment**: :broom: NIT -- loss of single element tuple changed
:shrug: Probably not an issue, as the `val` is already converted to a string by `esc_quotes(bytes | str)` but keep an eye out for edge-case regressions (albeit current tests _look_ unaffected) (Noted for completeness)

piplicenses.py
```diff
@@ -630,12 +624,12 @@ def esc_quotes(val: bytes | str) -> str:
...
         )
         lines.append(formatted_header)
         lines.extend(
             [
-                ",".join(['"%s"' % (esc_quotes(val),) for val in row])
+                ",".join([f'"{esc_quotes(val)}"' for val in row])
```
reactive-firewall, line 632: **Additional Comment**: :broom: NIT -- loss of single element tuple changed
:shrug: Probably not an issue, as the `val` is already converted to a string by `esc_quotes(bytes | str)` but keep an eye out for edge-case regressions (albeit current tests _look_ unaffected) (Noted for completeness)

piplicenses.py
```diff
@@ -811,23 +805,19 @@ def create_warn_string(args: CustomNamespace) -> str:
...
 
-    if args.with_license_file and not args.format_ == FormatArg.JSON:
+    if args.with_license_file and args.format_ != FormatArg.JSON:
         message = warn(
-            (
-                "Due to the length of these fields, this option is "
-                "best paired with --format=json."
-            )
+            "Due to the length of these fields, this option is "
+            "best paired with --format=json."
```
reactive-firewall, line 811: **Additional Comment**: :broom: NIT -- seems less readable IMHO
:shrug: Probably not an issue, but this change feels slightly less readable to me. (Noted for completeness)

piplicenses.py
```diff
@@ -811,23 +805,19 @@ def create_warn_string(args: CustomNamespace) -> str:
...
-            (
-                "When using this option, only --order=count or "
-                "--order=license has an effect for the --order "
-                "option. And using --with-authors and --with-urls "
-                "will be ignored."
-            )
+            "When using this option, only --order=count or "
+            "--order=license has an effect for the --order "
+            "option. And using --with-authors and --with-urls "
+            "will be ignored."
```
reactive-firewall, line 820: **Additional Comment**: :broom: NIT -- seems less readable IMHO
:shrug: Probably not an issue, but this change feels slightly less readable to me. (Noted for completeness)

piplicenses.py
```diff
@@ -929,16 +919,16 @@ def _verify_args(self, args: CustomNamespace) -> None:
...
         except LookupError:
             self.error(
-                "invalid code page '%s' given for '--filter-code-page, "
-                "check https://docs.python.org/3/library/codecs.html"
-                "#standard-encodings for valid code pages"
-                % args.filter_code_page
+                f"invalid code page '{args.filter_code_page}' given "
+                "for '--filter-code-page, check "
+                "https://docs.python.org/3/library/codecs.html#standard-encodings "  # noqa: E501
+                "for valid code pages"
```
reactive-firewall, line 925: :broom: NIT -- Consider resolving the `E501`

Consider splitting the anchor as before to avoid `E501` regression (line length limit)

E.g.,

```suggestion
                "for '--filter-code-page, check "
                "https://docs.python.org/3/library/codecs.html"
                "#standard-encodings for valid code pages"
```

test_piplicenses.py
```diff
@@ -896,7 +889,7 @@ def mocked_open(*args, **kwargs):
 
 def test_output_file_error(monkeypatch) -> None:
     def mocked_open(*args, **kwargs):
-        raise IOError
+        raise OSError
```
reactive-firewall, line 892: :broom: NIT -- consider improved replacement

please consider:
```suggestion
        raise OSError("Mock I/O OSError test-fixture.")
```

test_piplicenses.py
reactive-firewall: I'm fine with removing it if it does not change the test behavior. If it is needed by MyPy then I recommend we use a specific linter pragma (e.g., `# type: ignore[arg-type]` as needed).

test_piplicenses.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 # vim:fenc=utf-8 ff=unix ft=python ts=4 sw=4 sts=4 si et
```
reactive-firewall, line 1: :broom: NIT -- removing is also unnecessary

I'm inclined to disable Ruff rule `UP009` as there is no harm in [PEP 263](https://www.python.org/dev/peps/pep-0263/) style headers. I prefer the explicit header, but I'll accept the vim format line being enough, once updated with `ft=python3` (the rest is already correct):

```suggestion
# vim:fenc=utf-8 ff=unix ft=python3 ts=4 sw=4 sts=4 si et
```

and align to [PEP 3120](https://www.python.org/dev/peps/pep-3120/)

References:

- [PEP 263 - Defining a Source Code Encoding](https://www.python.org/dev/peps/pep-0263/)
- [PEP 3120 - Remove Implicit Unicode String Type](https://www.python.org/dev/peps/pep-3120/)

piplicenses.py
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
 # vim:fenc=utf-8 ff=unix ft=python ts=4 sw=4 sts=4 si et
```
reactive-firewall, line 2: :broom: NIT -- removing is also unnecessary

I'm inclined to disable Ruff rule `UP009` as there is no harm in [PEP 263](https://www.python.org/dev/peps/pep-0263/) style headers. I prefer the explicit header, but I'll accept the vim format line being enough, once updated with `ft=python3` (the rest is already correct):

```suggestion
# vim:fenc=utf-8 ff=unix ft=python3 ts=4 sw=4 sts=4 si et
```

and align to [PEP 3120](https://www.python.org/dev/peps/pep-3120/)

References:

- [PEP 263 - Defining a Source Code Encoding](https://www.python.org/dev/peps/pep-0263/)
- [PEP 3120 - Remove Implicit Unicode String Type](https://www.python.org/dev/peps/pep-3120/)

.github/workflows/python-package.yml
reactive-firewall, line 1: **Additional Comment**: Changes from rebase look fine.
  * For completeness, pinned hash remained `de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2`

.github/workflows/CD-PyPi.yml
reactive-firewall, line 1: **Additional Comment**: Changes from rebase look fine.
  * For completeness, pinned hash remained `de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2`

reactive-firewall (changes requested):
:+1: LGTM; with minor caveats.

Specifically, either revert the PEP-263 headers or update the vim directive to clarify python3 (and thus PEP-3120) to match.

See additional comments and suggestions inline.

piplicenses.py
DimitriPapadopoulos: I'm not sure about it. The `%` operator expects a tuple of arguments for string interpolation, but f-strings don't. So there's no actual loss or edge-case regression.

piplicenses.py
DimitriPapadopoulos: The string is already surrounded by the function call parentheses.

By the same logic, adding a 3rd pair of parentheses might make the code even more readable.

piplicenses.py
DimitriPapadopoulos: The PEP8 limit of 79 characters is considered too small nowadays and most recent linters extend that limit to 88 by default. The E501 error would disappear with the default settings of recent linters and formatters such as _Black_.

In that context, splitting the URL for the sake of old times' sake doesn't make it more readable.

test_piplicenses.py
DimitriPapadopoulos: Agreed. However, my idea of this PR was to apply Ruff rules, not improve error messages. I would rather improve error messages in a different PR.

test_piplicenses.py
DimitriPapadopoulos: No one uses `# -*- coding: utf-8 -*-` nowadays: Python files  are UTF-8 by default, and I suspect `python` **is** Python 3. I cannot find the documentation for `ft=python` in Vim. Any clue?

DimitriPapadopoulos:
> I'm fine with removing it if it does not change the test behavior. If it is needed by MyPy then I recommend we use a specific linter pragma (e.g., # type: ignore[arg-type] as needed).

Tests pass, so it is not needed by MyPy, is it?

piplicenses.py
reactive-firewall: > By the same logic, adding a 3rd pair of parentheses might make the code even more readable.

LOL. Fair enough.

test_piplicenses.py
reactive-firewall: :thinking: indeed; a separate focused PR is probably a good idea.

test_piplicenses.py
reactive-firewall: I generally agree that `# -*- coding: utf-8 -*-` is very archaic these days (e.g., post 2020 EoL of python2), and fully admit that this is merely a nitpick point.

After checking some more, looks like both are file type plugins for vim based editors, with the archaic `ft=python` (via `ft-python-plugin`) being more commonly available (see [online vim help : filetype](https://vimhelp.org/filetype.txt.html#filetype.txt)), further it only seems to configure spacing (I'm unsure it even handles syntax highlighting)

With that in mind, I'm fine leaving the vim directives as they are. I'll go-ahead and approve the PR as is.

PR #268, opened 2026-01-18
Run black 26.1.0

### LINKED ISSUE #269: Apply Black 26.1.0 to source files
_Black_ [26.1.0](https://github.com/psf/black/releases/tag/26.1.0) introduces the 2026 stable style, implying a few minor changes.

You need to rerun this version of _Black_ to fix CI.

Probably apply to branch [dev-5.5.1](https://github.com/raimon49/pip-licenses/tree/dev-5.5.1) to fix CI in that branch.

DimitriPapadopoulos:
_Black_ [26.1.0](https://github.com/psf/black/releases/tag/26.1.0) introduces the 2026 stable style, implying a few minor changes.

Fixes #269.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -26,6 +26,7 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 SOFTWARE.
 """
+
 from __future__ import annotations
 
 import argparse
@@ -44,7 +45,6 @@ from pathlib import Path
 from typing import TYPE_CHECKING, cast
 
 from prettytable import HRuleStyle, PrettyTable, RowType
-
 
 if sys.version_info >= (3, 11):
     import tomllib
@@ -317,10 +317,10 @@ def get_packages(
         return (included_file, included_text)
 
     def get_pkg_info(pkg: Distribution) -> dict[str, str | list[str]]:
-        (license_file, license_text) = get_pkg_included_file(
+        license_file, license_text = get_pkg_included_file(
             pkg, "LICEN[CS]E.*|COPYING.*"
         )
-        (notice_file, notice_text) = get_pkg_included_file(pkg, "NOTICE.*")
+        notice_file, notice_text = get_pkg_included_file(pkg, "NOTICE.*")
         pkg_info: dict[str, str | list[str]] = {
             "name": pkg.metadata["name"],
             "version": pkg.version,


DimitriPapadopoulos:
The CI error is because _Black_ is run with Python 3.9, but [26.1.0](https://github.com/psf/black/releases/tag/26.1.0) requires Python 3.10 (https://github.com/psf/black/pull/4842).

I don't understand why lint with _Black_  for each tested version of Python. A single version of Python is good enough. Same for MyPy.

DimitriPapadopoulos:
You should merge this into [dev-5.5.1](https://github.com/raimon49/pip-licenses/tree/dev-5.5.1) instead of [master](https://github.com/raimon49/pip-licenses/tree/master).

piplicenses.py
```diff
@@ -45,7 +46,6 @@
 
 from prettytable import HRuleStyle, PrettyTable, RowType
 
-
```
reactive-firewall, line 48: :broom: NIT - Consider leaving the double newlines.

Rationale: 
:thinking: unless black is insisting on removing the double blanks, I prefer the style of double newline outside of class member definitions to visually highlight the gaps in global scope.

```suggestion


```

piplicenses.py
```diff
@@ -317,10 +317,10 @@ def get_pkg_included_file(
         return (included_file, included_text)
 
     def get_pkg_info(pkg: Distribution) -> dict[str, str | list[str]]:
-        (license_file, license_text) = get_pkg_included_file(
+        license_file, license_text = get_pkg_included_file(
             pkg, "LICEN[CS]E.*|COPYING.*"
         )
-        (notice_file, notice_text) = get_pkg_included_file(pkg, "NOTICE.*")
+        notice_file, notice_text = get_pkg_included_file(pkg, "NOTICE.*")
```
reactive-firewall, line 323: **Additional Comments**:
  This is the refactor black insists on.
  :+1: Specifically, the refactoring of tuple assignments for strings is approved!

reactive-firewall:
> I don't understand why lint with _Black_ for each tested version of Python. A single version of Python is good enough. Same for MyPy.

Regarding `black`, I totally agree, it is certainly opinionated enough to be run once on the latest stable version of python. It currently flags when it would re-write code, so as a hack :speak_no_evil: it is used to abort recording test-coverage on failure. This is not a requirement as far as I understand though. I've been considering migrating to something else. E.g.,

CI_Alt_idea

Regarding `MyPy`:
`MyPy` could _theoretically_ differ between pythons (albeit not in our case), further as I am not a fan of MyPy, I'd even be fine removing it, of-course only after discussing with the community (assuming they don't object :shrug:). It is historically tied-in with the testing logic that pre-dates my activity with the project, so I've left it alone. I do intend to keep the code annotated with typing after-all.

reactive-firewall:
@DimitriPapadopoulos Heads up, I'll merge this after I familiarize myself with "the 2026 stable style" introduced by black 26.1.0.

:bow: Again Thank you for your efforts on this PR.

piplicenses.py
DimitriPapadopoulos: These are not class member definitions, but imports. That is probably why Black removed the extra line.

With that said, either you're willing to run Black on the codebase, or not. There's no middle way! The Ruff formatter provides more options, although I suspect it will be as opinionated as Black in this specific case.

DimitriPapadopoulos:
There's no need to familiarize with "the 2026 stable style". The whole point of formatters is to run them and forget about code formatting.

DimitriPapadopoulos:
You could also run Black **before** instead of in parallel with cross-Python testing. Alternatively, GitHub workflows should allow to abort cross-Python testing if Black fails.

As for MyPy, the **target version of Python** is the same as the version of Python used to execute MyPy by default. However, you may set the target version to be different from the version of Python used to execute MyPy. See [`--python-version`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-python-version). 

You just need to run Mypy once with the earliest version of Python you support as the target.

DimitriPapadopoulos:
> Regarding `black`, I totally agree, it is certainly opinionated enough to be run once on the latest stable version of python. It currently flags when it would re-write code, so as a hack 🙊 it is used to abort recording test-coverage on failure. This is not a requirement as far as I understand though. I've been considering migrating to something else. E.g.,

See #271, but I'm not sure my merge request implements "Linting" **in parallel**, since the `build` job depends on the new `lint` job:
```yaml
  build:
    runs-on: ubuntu-latest
    needs: lint

```
I've done that because the existing situation makes sense to me: do not start extended testing unless basic linting tests do pass.

piplicenses.py
reactive-firewall: > With that said, either you're willing to run Black on the codebase, or not. There's no middle way! The Ruff formatter provides more options, although I suspect it will be as opinionated as Black in this specific case.

:+1: fair enough

PR #271, opened 2026-01-21
Run Black and MyPy only once

### LINKED ISSUE #270: Run "linters" once, with a single version of Python
_Black_ is run multiple times, for each supported version of Python:
https://github.com/raimon49/pip-licenses/blob/6beaf9cdc7e223bb6ca7911798f1dced8cfbb1f6/.github/workflows/python-package.yml#L15-L31

* This is not efficient, the version of Python used to run _Black_ is irrelevant.
* The current version of _Black_ requires Python 3.10, it cannot run under Python 3.9.

Same for MyPy.

Probably apply to branch [dev-5.5.1](https://github.com/raimon49/pip-licenses/tree/dev-5.5.1) to fix CI in that branch.

DimitriPapadopoulos:
* Use the latest stable version of Python to execute Black and MyPy.
* Define Python 3.9 (earliest supported version) as the target for MyPy.

Fixes #270.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -49,7 +49,7 @@ from prettytable import HRuleStyle, PrettyTable, RowType
 if sys.version_info >= (3, 11):
     import tomllib
 else:
-    import tomli as tomllib
+    import tomli as tomllib  # type: ignore[import-not-found]
 
 if TYPE_CHECKING:  # pragma: no cover
     from email.message import Message


DimitriPapadopoulos:
An alternative would be to run linting tools Black/Ruff/Mypy using [pre-commit.ci](https://pre-commit.ci/) instead of GitHub workflows. That's what lots of Python projects do nowadays.

Benefits:
* Linting is easily run locally on developers' machines and during CI, producing the exact same output.
*  [pre-commit.ci](https://pre-commit.ci/) automatically applies straightforward fixes to PRs (in the form of an additional commit in the PR), leaving as CI errors only issue with unsafe fixes or without fixes that require developer action.

Caveats:
* Adds a CI operator. It mostly means an extra organisation to trust in the software assembly line, in addition to Microsoft.
* The lead developer of pre-commit is opinionated and his interactions with the community are sometimes perceived as "brutal".

.github/workflows/python-package.yml
```diff
@@ -10,12 +10,35 @@ on:
...
+        with:
+          options: "--check --diff"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
+      - name: Validate static typing with mypy
+        run: | # the target Python version should be the earliest supported version
+          mypy --install-types --non-interactive --python-version 3.9 .
+
```
reactive-firewall, line 35: * Let's keep the python version specific, and singular (per-job at-lease), and use either a [job env variable](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idenv) or use a [step id](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsid) for `actions/setup-python` and use it's outputs (see [actions/setup-python outputs](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#outputs))

  * **EDIT**: let's target python python 3.14 (latest stable) ~3.10~ for now, as 3.9 is just for support of macOS vendor provide python compatibility
  * use either `CI_PYTHON` (env name) or `python` (id value)
  
Consider:

```suggestion
  lint:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout repository
        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
        with:
          persist-credentials: false
      - name: Set up Python
        id: python
        uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6.1.0
        with:
          python-version: '3.10'  # black GHA fails before v3.10
      - name: Format with black
        id: black
        uses: psf/black@stable
        with:
          options: "--check --diff"
      - name: Install dependencies
        id: deps
        if: ${{ !cancelled() }}
        run: |
          python -m pip install --upgrade pip
          if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
      - name: Validate static typing with mypy
        id: mighty
        if: ${{ !cancelled() }}
        run: | # the target Python version should be the earliest supported version
          mypy --install-types --non-interactive --python-version ${CI_PYTHON} .
        env:
          CI_PYTHON: '${{ steps.python.outputs.python-version }}'  # security mitigation: expand GHA context before use in shell.

```

<details><summary>Alternatively consider</b></summary>

```suggestion
  lint:
    runs-on: ubuntu-latest
    env:
      CI_PYTHON: '3.10'  # controls lint job python version
    steps:
      - name: Checkout repository
        uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
        with:
          persist-credentials: false
      - name: Set up Python
        uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6.1.0
        with:
          python-version: '${{ env.CI_PYTHON }}'
      - name: Format with black
        uses: psf/black@stable
        with:
          options: "--check --diff"
      - name: Install dependencies
        run: |
          python -m pip install --upgrade pip
          if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
      - name: Validate static typing with mypy
        run: | # the target Python version should be the earliest supported version
          mypy --install-types --non-interactive --python-version ${CI_PYTHON} .

```

</details>

DimitriPapadopoulos: I wouldn't do that:
* While running a linter with Python 3.10 or any other version supported by the linter is OK, I think the **target version** of Python should be the earliest version of Python supported by the project - hence `--python-version 3.9` to define the target version to be 3.9.
* There's no need to run linters such as MyPy using a specific version of Python. Recent versions of linters will work out of the box with recent versions of Python. "The majority" of projects I know of use `python-version: '3.x'` to run developer tools and I've never heard of failures.

DimitriPapadopoulos: For example:

[**pip/.github/workflows/ci.yml**](https://github.com/pypa/pip/blob/e9dc2fbee0198254a7373eb1d9d47beac2a5323d/.github/workflows/ci.yml#L32-L36)
Lines 32 to 36 in [`e9dc2fb`](https://github.com/pypa/pip/commit/e9dc2fbee0198254a7373eb1d9d47beac2a5323d)
```yaml
      - uses: actions/setup-python@v6
        with:
          python-version: "3.x"
      - run: pip install nox
      - run: nox -s docs
```

[**pip/.github/workflows/ci.yml**](https://github.com/pypa/pip/blob/e9dc2fbee0198254a7373eb1d9d47beac2a5323d/.github/workflows/ci.yml#L74-L85)
Lines 74 to 85 in [`e9dc2fb`](https://github.com/pypa/pip/commit/e9dc2fbee0198254a7373eb1d9d47beac2a5323d)
```yaml
      - uses: actions/setup-python@v6
        with:
          python-version: "3.x"
      - name: Set up git credentials
        run: |
          git config --global user.email "pypa-dev@googlegroups.com"
          git config --global user.name "pip"

      - run: pip install nox
      - run: nox -s prepare-release -- 99.9
      - run: nox -s build-release -- 99.9
      - run: pipx run check-sdist
```

reactive-firewall: > I wouldn't do that:
> 
> * While running a linter with Python 3.10 or any other version supported by the linter is OK, I think the **target version** of Python should be the earliest version of Python supported by the project - hence `--python-version 3.9` to define the target version to be 3.9.

Alright, I'm convinced. Specifically the `--python-version 3.9` should stay.

However, (call it an abundance of caution :shrug:) I still insist that the GHA job be configured to target a _pinned_ version of python (or even a specific range like `3.9 thru 3.14`, but that somewhat defeats the gains of this PR), specifically for `python-setup`, although it need not be implemented as elaborate as I suggested on account of `mypy`'s `--python-version` argument.

---

> * There's no need to run linters such as MyPy using a specific version of Python.

I respectfully disagree that there are no reasons.

Please kindly consider:
  * build reproducibility (version granularity)
    * and scouring the logs instead of the config is often a waste of effort.
  * supply-chain security (version pinning)
    * it's the form of `3.x` at issue, not the principle
  * :broom: NIT -- I have personally seen GHA runners' environment just have errors (e.g., during GH rolling out ubuntu 24 lead to lots of inconsistencies in "latest" versions)
  * :broom: NIT -- The adage: _"we are not shipping"_ those other projects applies (let's )

DimitriPapadopoulos:
The version of Python used to run the linter is hardy a question of build reproducibility. However, I can understand you would like to pin versions nevertheless. I will specify `3.14`, but note that patch versions will still change.

DimitriPapadopoulos:
I still need to fix this error:
```
piplicenses.py:52: error: Cannot find implementation or library stub for module named "tomli"  [import-not-found]
piplicenses.py:52: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Installing missing stub packages:
/opt/hostedtoolcache/Python/3.14.2/x64/bin/python -m pip install types-Pygments types-docutils types-pexpect
```

DimitriPapadopoulos:
Looks like a MyPy bug/limitation: the target is Python 3.9, so here it attempts to import `tomli`:
```python
if sys.version_info >= (3, 11):
    import tomllib
else:
    import tomli as tomllib
```
But then `tomli` is not available since this is Python 3.14. So MyPy needs `tomli` although `tomllib` is available in the current Python version. 

See https://github.com/python/mypy/issues/13914.

Not sure how to work around it:
* Add `tomli` to `dev-requirements.txt` (how)?
* Add `# type: ignore[import-not-found]`?

I'm going for the second, since this looks like a MyPy issue.

.github/workflows/python-package.yml
reactive-firewall, line 1: ## :sparkles: Resolved!

:ballot_box_with_check: Specifically, the version pinning is approved!

<details><summary>:page_facing_up: Additional Comments from review</summary>
 * :broom: NIT -- Personal style differences aside, all concerns have been discussed and requested changes were applied. These changes are now precise, and look good as they are; @DimitriPapadopoulos, your time and efforts on this PR are greatly appreciated! :bow: Thank you.

</details>

piplicenses.py
```diff
@@ -49,7 +49,7 @@
 if sys.version_info >= (3, 11):
     import tomllib
 else:
-    import tomli as tomllib
+    import tomli as tomllib  # type: ignore[import-not-found]
```
reactive-firewall, line 53: > See [python/mypy#13914](https://github.com/python/mypy/issues/13914).
> 
> Not sure how to work around it:
> 
> * Add `tomli` to `dev-requirements.txt` (how)?
> * Add `# type: ignore[import-not-found]`?
> 
> I'm going for the second, since this looks like a MyPy issue.

:star_struck: Prefect! The workaround is approved as is!

PR #286, opened 2026-01-29
Use Ruff as linter and formatter

### LINKED ISSUE #273: Black → Ruff formatter?
How about moving from Black to the Ruff formatter?

* I would like to add a linting step based on Ruff, to apply Ruff rules such as those in #261.
* Ruff is both a formatter like Black and a linter applying rules from a multitude of existing linters. Using Ruff as a linter and a formatter would reduce the number of developer tools.
* Ruff is much faster than Black, but then Black is not the "linting" bottleneck, that would be MyPy.

DimitriPapadopoulos:
- Enforce Ruff rules from #261.
- Replace isort with the Ruff linter and [isort (I)](https://docs.astral.sh/ruff/rules/#isort-i) rules. Note that isort was previously **not** run in CI which means isort rules had actually **not** been applied.
- Replace Black with the Ruff formatter. I have not tried to adjust the default Ruff rules to obtain the exact same formatting as Black, as I find Ruff formatting to be actually more readable.

This reduces the number of tools from 4 (black, isort, pycodestyle, mypy), or even 5 (autopep8, black, isort, pycodestyle, mypy), to 2 (ruff, mypy).

Fixes #273. I understand this might not be accepted, but I thought I'd give a try, to at least showcase the benefit of replacing multiple linting/formatting tools with a single one.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -21,17 +21,14 @@ jobs:
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: '3.14'
-      - name: Format with black
-        uses: psf/black@stable
-        with:
-          options: "--check --diff"
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
           if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
-      - name: Lint with ruff
+      - name: Lint and format with ruff
         run: |
           ruff check --diff
+          ruff format --diff
       - name: Validate static typing with mypy
         run: | # the target Python version should be the earliest supported version
           mypy --install-types --non-interactive --python-version 3.9 .
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ help:
 	@echo '    local-install   Install locally'
 	@echo '    local-uninstall Uninstall locally'
 	@echo '    update-depends  Re-compile requirements for development'
-	@echo '    lint            Re-lint by black, ruff and mypy'
+	@echo '    lint            Re-lint by ruff and mypy'
 	@echo '    test            Run unittests'
 	@echo '    deploy          Release to PyPI server'
 	@echo '    test-deploy     Release to Test PyPI server'
@@ -39,7 +39,7 @@ update-depends:
 .PHONY: lint
 lint:
 	ruff check
-	black .
+	ruff format
 	mypy --install-types --non-interactive .
 
 .PHONY: test
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,6 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    "black",
     "docutils",
     "wheel",
     "mypy",
@@ -65,9 +64,6 @@ readme = {file = ["README.md", "CHANGELOG.md"], content-type = "text/markdown"}
 
 [tool.pytest.ini_options]
 addopts = "-v --cov --cov-report term-missing"
-
-[tool.black]
-line-length = 79
 
 [tool.ruff]
 line-length = 79
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -6,8 +6,6 @@
 #
 backports-tarfile==1.2.0
     # via jaraco-context
-black==25.11.0
-    # via pip-licenses (pyproject.toml)
 build==1.4.0
     # via pip-tools
 certifi==2026.1.4
@@ -17,9 +15,7 @@ cffi==2.0.0
 charset-normalizer==3.4.4
     # via requests
 click==8.1.8
-    # via
-    #   black
-    #   pip-tools
+    # via pip-tools
 coverage[toml]==7.10.7
     # via pytest-cov
 cryptography==46.0.4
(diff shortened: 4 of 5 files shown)


pyproject.toml
```diff
@@ -66,15 +63,38 @@ version = {attr = "piplicenses.__version__"}
 readme = {file = ["README.md", "CHANGELOG.md"], content-type = "text/markdown"}
 
 [tool.pytest.ini_options]
-addopts = "--pycodestyle -v --cov --cov-report term-missing"
+addopts = "-v --cov --cov-report term-missing"
```
reactive-firewall, line 66: :bulb: let's take a look at [pytest-ruff](https://pypi.org/project/pytest-ruff/) as the replacement for `pycodestyle` to ensure closer equivalence in the testing workflow (e.g., `make test` still does some format checks)

pyproject.toml
```diff
@@ -66,15 +63,38 @@ version = {attr = "piplicenses.__version__"}
...
+    "I",      # isort
+    "PERF",   # Perflint
+    "PGH",    # pygrep-hooks
+    "UP",     # pyupgrade
+    "FURB",   # refurb
+    "RUF",    # Ruff-specific rules
+]
+ignore = [
+    "C401",   # Unnecessary generator (rewrite as a set comprehension)
+    "F841",   # Local variable is assigned to but never used
```
reactive-firewall: :broom: NIT -- Let's open a ticket to clean these up, and call it technical debt.

* Specifically `F841`, nor `C401`, nor `RUF021`, need be ignored.

Makefile
```diff
@@ -38,8 +38,8 @@ update-depends:
 
 .PHONY: lint
 lint:
-	black .
-	isort .
+	ruff check
+	ruff format
```
reactive-firewall, line 42: :broom: NIT -- for consistency consider preserving the path argument

Please consider:
```suggestion
	ruff check .
	ruff format .
```

pyproject.toml
```diff
@@ -35,17 +35,14 @@ dependencies = [
...
-    "autopep8",
-    "black",
     "docutils",
-    "isort",
     "wheel",
     "mypy",
     "pip-tools",
     "pypandoc",
     "pytest-cov",
-    "pytest-pycodestyle",
```
reactive-firewall, line 47: Consider:

```suggestion
    "pytest-ruff",
```

.github/workflows/python-package.yml
```diff
@@ -54,7 +54,7 @@ jobs:
           if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
       - name: Test with pytest
         run: |
-          pytest --pycodestyle -v --cov --cov-report=xml
+          pytest -v --cov --cov-report=xml
```
reactive-firewall, line 57: Please Consider:
```suggestion
          pytest --ruff --ruff-format -v --cov --cov-report=xml
```

See related suggestions.

Makefile
DimitriPapadopoulos: Black requires the argument, while Ruff doesn't. I could add the argument, but most projects use it without the argument, and its documentation suggests using it without the argument too:
https://docs.astral.sh/ruff

DimitriPapadopoulos:
I have to admit, removing pytest-pycodestyle rather belongs to #271, which should have been "Run Black,pycodestyle and MyPy only once" instead of "Run Black and MyPy only once".

But it's never too late. The rationale:
* I think we should avoid running pycodestyle/ruff with each version of Python test in CI. We already have a dependency of tests on linting in CI, so linting **will** be performed before testing:
  https://github.com/raimon49/pip-licenses/blob/d29ccddaf92d36996a93a8c76af53c052acfe8db/.github/workflows/python-package.yml#L36-L38
* We could add the same dependency to the Makefile:
  https://github.com/raimon49/pip-licenses/blob/d29ccddaf92d36996a93a8c76af53c052acfe8db/Makefile#L45-L47
  but then that's not what the current documentation suggests:
  https://github.com/raimon49/pip-licenses/blob/d29ccddaf92d36996a93a8c76af53c052acfe8db/Makefile#L15-L16

pyproject.toml
DimitriPapadopoulos: That would apply not only to `make test`, but also to CI, wouldn't it? I think we want to avoid that, at least in CI. See below.

reactive-firewall:
> * I think we should avoid running pycodestyle/ruff with each version of Python test in CI. We already have a dependency of tests on linting in CI, so linting **will** be performed before testing:

* :thinking: Good point. Let's consider this part of GHI #271 (ruff --ruff-format stuff is indeed part of linting)

---

> but then that's not what the current documentation suggests

* Makefile documentation will need to be updated via GHI #290

* in general, I'll refer back to my intended inter-job-dependencies of CI from:
  CI_Alt_idea

---

> We could add the same dependency to the Makefile

* Also consider this idea part of GHI #290

* this is the route I actually prefer, however, we need to avoid possible regression of the recent work on separating out linting from cross-python test coverage...
  * We could just consider something like a new `PHONY` make target. E.g.,
  ```Makefile
  .PHONY: local-ci-check
  local-ci-check:: build lint test
  	@true  # does nothing successfully (if reached)
  
  ```
  * the name is just arbitrary, based on the rational there is already a `local-install` target :shrug:

PR #287, opened 2026-01-29
Simplify PR-221 - Improve stability of multi-lined allow/fail arg lists

reactive-firewall:
* Supersedes PR #221 by @anmol-aidora 

> When setting the `allow-only` and `fail-on` arguments, the list can get very long very fast, especially when switching `partial` to `false`. There is a need to make sure multi-line arguments work properly.
> 
> Currently when passing multi-line allow-only and fail-on arguments in `pyproject.toml`, the `fail_on_licenses` and `allow_only_licenses` sets were found to contain the empty string `''`.
> 
> e.g.
> 
> ```
> allow-only="""
> MIT;Apache;
> """
> ```
> 
> sets allow_only_licenses to:
> 
> ```
> {'', 'Apache', 'MIT'}
> ```
> 
> The code checks for the same and removes it from the sets, such that there is no empty string:
> 
> ```
> {'Apache', 'MIT'}
> ```

---

* refactored to just add `filter(None, ...)` by @reactive-firewall 
* Includes additional unit-test

### DIFF
--- a/test_piplicenses.py
+++ b/test_piplicenses.py
@@ -970,6 +970,53 @@ def test_allow_only_partial(monkeypatch) -> None:
     )
 
 
+def test_allow_only_with_empty_tokens(monkeypatch) -> None:
+    # same as test_allow_only but with extra semicolons/whitespace
+    licenses = (
+        "Bsd License",
+        "Apache Software License",
+        "Mozilla Public License 2.0 (MPL 2.0)",
+        "Python Software Foundation License",
+        "Public Domain",
+        "GNU General Public License (GPL)",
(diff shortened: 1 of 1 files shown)


test_piplicenses.py
reactive-firewall, line 1: :broom: NIT -- Might need to refactor if using ruff

* not checked with `ruff` for linting; passes black

reactive-firewall (commented):
## :+1: LGTM;

> [!NOTE]
> with caveat: I am the project maintainer/admin and was this PR's author and I self-reviewed.

PR #298, opened 2026-02-14
Lint using pre-commit - locally and in CI

### LINKED ISSUE #288: Use pre-commit.ci to run linter during CI?
The idea is to run linting and formatting tools using [pre-commit.ci](https://pre-commit.ci/) instead of GitHub workflows.

Tests will still be run from GitHub workflows.

Benefits:
- Run the exact same linting tools either locally on the developers' machines or during CI.
- Let [pre-commit.ci](https://pre-commit.ci/) add a commit to any PR to automatically format and fix those linter issues that are safe to fix automatically. This additional commit is then reviewed like the rest.

Caveats:
- Dependency on an new operator, in addition to GitHub.

DimitriPapadopoulos:
Run pre-commit as a GitHub Action instead of giving pre-commit.ci access to this repository.

Fixes #288.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -21,17 +21,8 @@ jobs:
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
           python-version: '3.14'
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
-      - name: Lint and format with ruff
-        run: |
-          ruff check --diff
-          ruff format --diff
-      - name: Validate static typing with mypy
-        run: | # the target Python version should be the earliest supported version
-          mypy --install-types --non-interactive --python-version 3.9 .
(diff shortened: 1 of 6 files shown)


reactive-firewall (commented):
## :no_good: Un-necessary risk remains

While the implementation looks fine for pre-commit, I still find the risk of granting write permissions to go against the security BP of "least privileged". I appreciate the work and time you've put into this patch. Thank you.

**Additional Comments**:

After looking more into `pre-commit` and `prek` (as discussed in GHI #288) I have decided that `prek` is preferred¹.
This PR will thus be included² & superseded by PR #313.

While it was discussed that, we might grant write access to CI to allow auto formatting, I found both `pre-commit` and `prek` do not require git-hook integration to use, allowing for the flexibility between use in CI (read-only) and use locally by developers (read/write).

> [!IMPORTANT]
> Closing as "Superseded"

---

TL;DR
_1: `prek` is easier to invoke than pre-commit when not installed as git hooks, and typically runs faster than pre-commit, and allows proper ordering of checks_
_2: including some work, to preserve your efforts, but heavily refactored for reduced security surface and re-tooling._

PR #301, opened 2026-02-15
Use f-strings

DimitriPapadopoulos:
**EDIT**: The main focus of this PR is on enhancing code readability and modernizing string formatting by utilizing f-strings.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -1060,11 +1060,12 @@ def create_parser(
     format_options = parser.add_argument_group("Format options")
     verify_options = parser.add_argument_group("Verify options")
 
+    lit_prog_pat = "%(prog)s"
     parser.add_argument(
         "-v",
         "--version",
         action="version",
-        version=f"%(prog)s {__version__}",  # type: ignore [name-defined]
+        version=f"{lit_prog_pat} {__version__}",
     )
 
     common_options.add_argument(


piplicenses.py
```diff
@@ -195,7 +195,7 @@ def normalize_version(version_string):
         str: A normalized version string in PEP 440 format or empty if invalid.
     """
     _regex = re.compile(
-        r"^\s*" + VERSION_PATTERN + r"\s*$",
+        rf"^\s*{VERSION_PATTERN}\s*$",
```
reactive-firewall, line 198: **Additional Comment**:
:star_struck: Ok that's a new trick for me. Makes a lot of sense in hindsight though.

piplicenses.py
```diff
@@ -1060,7 +1061,10 @@ def create_parser(
     verify_options = parser.add_argument_group("Verify options")
 
     parser.add_argument(
-        "-v", "--version", action="version", version="%(prog)s " + __version__
+        "-v",
+        "--version",
+        action="version",
+        version=f"%(prog)s {__version__}",  # type: ignore [name-defined]
     )
```
reactive-firewall, line 1069: :broom: NIT -- :roll_eyes: `mypy` doesn't really understand python.

This was originally concatenated for `mypy`. The idea then was that concatenation was better than the type exception.

<details><summary>TL;DR - This is a good example of where `mypy` is just wrong.</summary>

Argparse handles the `%(prog)s` dynamicly based on the launch-time name of the program, E.g.,
This works correctly for: `./venv/pip-licenses/bin/python3 -B ./piplicenses.py --version`

```console
$ make setup >/dev/null # setup venv
$ ./venv/pip-licenses/bin/python3 -B ./piplicenses.py --version
piplicenses.py 5.5.2
```

**AND** this works correctly for (if installed in shell's `PATH`): `pip-licenses --version`

```console
$ pip-licenses --version
pip-licenses 5.5.2
```

**HOWEVER** _`mypy`_ doesn't understand that `argparse` (a python builtin now days) will do this behavior correctly and expects the fstring to mean something  other than it does (e.g., the literal `argparse` pattern like it does, ... `f"%(prog)s {__version}"` → `"%(prog)s 5.5.2"`)

</details>

Consider defining the literal `argparse` pattern as a variable:

```suggestion
    lit_prog_pat: str = "%(prog)s"
    parser.add_argument(
        "-v",
        "--version",
        action="version",
        version=f"{lit_prog_pat} {__version__}",
    )
```

`mypy` should accept this and the dynamic function should still be preserved with this fstring.

PR #299, opened 2026-02-14
PEP 735 compliance

DimitriPapadopoulos:
# Patch Notes

This pull request migrates the project's dependency management to comply with [PEP-735](https://peps.python.org/pep-0735/) by refactoring the `pyproject.toml` file. It replaces the deprecated `[project.optional-dependencies]` section with the new standardized `[dependency-groups]` format, enhancing clarity and organization for managing development dependencies.

## Comprehensive Table of Package Managers Supporting PEP 735

| Package Manager | PEP 735 Support | Notes | Documentation Link |
|------------------|-----------------|-------|---------------------|
| **Pip**          | Yes             | Officially supports dependency groups since version **25.1**. Allows installation using the `--dependency-groups` option. | [Pip Documentation](https://pip.pypa.io/en/stable/user_guide/#dependency-groups) |
| **Pip-tools**    | Planned         | There are ongoing discussions and an open issue (#2062) for implementing PEP 735 support in pip-tools. | [Pip-tools GitHub Issue](https://github.com/jazzband/pip-tools/issues/2062) |
| **Poetry**       | Planned         | Discussions for implementing support are ongoing, with expected future updates aligning with PEP 735 standards. | [Poetry GitHub Issue](https://github.com/python-poetry/poetry/issues/3886) |
| **Tox**          | In Development    | Current proposals suggest integration with dependency groups, but not yet implemented. | [Tox GitHub Issue](https://github.com/tox-dev/tox/issues/1574) |
| **ReadTheDocs**  | In Discussion   | Proposals are underway to incorporate support for dependency groups into documentation builds. | [ReadTheDocs Discussion](https://github.com/readthedocs/readthedocs.org/issues/6798) |
| **UV**           | Yes             | Added support in version **0.4.27**, allowing users to utilize dependency groups effectively. | [UV Documentation](https://uv.readthedocs.io/en/latest/) |
| **Easy_install** | No              | Considered deprecated; no current support for PEP 735 due to migration to more modern tools. | [Setuptools Documentation](https://setuptools.pypa.io/en/latest/easy_install.html) |
| **Homebrew**     | No              | Primarily focuses on system-level package management; does not support Python's PEP 735 specifications. | [Homebrew Documentation](https://docs.brew.sh/) |
| **apt-get**      | No              | Designed for managing system packages for Linux, and does not include support for Python-specific PEP 735 group mechanisms. | [Apt Documentation](https://manpages.ubuntu.com/manpages/bionic/man8/apt-get.8.html) |

[...]

PR #303, opened 2026-02-15
pip-tools, virtualenv → uv

DimitriPapadopoulos:
Supersedes #299.


reactive-firewall:
## :no_good: Early Feedback on Your Draft PR

### Points to Address Before Review:

1. **Maintenance Considerations for `uv`:**
   - ~I understand the appeal of `uv`, but I'm hesitant to maintain it in the 5.x series. I can~ I have open a `dev-v6.0` branch for now if you want to proceed with this integration.
   - I have concerns about `uv` storing two sets of dependencies in one lock file:
     - The `source` configuration is set to the same top-level value: "https://pypi.org/simple" (PEP 503). It might be beneficial to consider using a more specific source URL like `https://pypi.org/simple/<project>` or a hashed version `/simple/<project>/@<hash-type>=<hash>`.
     - The configuration file for `uv` seems to be duplicating or caching metadata that could be retrieved from the source repository (PEP 503 + PEP 691), which may add to our maintenance load.
   - Since I don't personally use `uv` in my projects, this adds some extra maintenance overhead that I’d prefer to avoid:
     - We'll need to update the Docker components.
     - The Dependabot configuration (see [Dependabot Options Reference](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference)) will also need adjustments to support this integration.
     - ~**NIT:** I’ll need to familiarize myself with `uv` to provide a thorough review and maintain these changes effectively, but I’m here to support you in that.~ (this is mostly done)

2. **Focus on `pip-licenses`:**
   - Remember that the project is (and will remain) `pip-licenses`, with `pip` being intentional in the name.
   - While `uv` does help navigate around some issues with `pip-tools`, I have reservations (refer to my first point).
   - We will need to ensure robust functional testing for both package managers (`pip` and `uv`) to avoid regressions. This could include setting up a separate GitHub Workflow, e.g., `.github/workflows/python-managers.yml`, with distinct jobs for testing `uv` and `pip`, ensuring that both **production** installs work seamlessly across platforms. A separate PR for this may also be worthwhile due to the added complexity and the `pip-tools` blockers.

---

I appreciate your efforts on this PR and hope this feedback helps streamline the process moving forward. Please feel free to reach out if you'd like to discuss any of these points further!

DimitriPapadopoulos:
It's just that:
- `pip-compile` is slow and plagued by bugs that prevent it from doing well the few things it's supposed to do. `uv` is much faster and appears to be doing better. That's the ~~[`pip-compile` bug](https://github.com/DimitriPapadopoulos/pip-licenses/blob/e84e09f181d0cb39218c762eec9432b7ffd31bc5/pyproject.toml#L47-L50)~~ virtualenv bug that prompted me to move to `uv`: 
  ```
      # incompatibility with Python 3.9 because of a pip-compile bug:
      #   "filelock>=3.16.1,<4; python_version<'3.10'",
      #   "filelock>=3.20.1,<4; python_version>='3.10'",
      "virtualenv<20.36.0",
  ```
- Generally speaking Python is showing its age, compared to languages like Go or Rust. It has evolved from a simple to a moderately complex language, with a much steeper learning curve for beginners. More importantly, its ecosystem is a complete dog's breakfast with a multitude of tools for the same task. Tools like `ruff` or `uv` could bring some order in that mess.

DimitriPapadopoulos:
I agree, such a change might not belong to the 5.x series.

reactive-firewall:
> It's just that:
> 
> That's the [`pip-compile` bug](https://github.com/DimitriPapadopoulos/pip-licenses/blob/e84e09f181d0cb39218c762eec9432b7ffd31bc5/pyproject.toml#L47-L50) that prompted me to move to `uv`:
>   ```
>       # incompatibility with Python 3.9 because of a pip-compile bug:
>       #   "filelock>=3.16.1,<4; python_version<'3.10'",
>       #   "filelock>=3.20.1,<4; python_version>='3.10'",
>       "virtualenv<20.36.0",
>   ```

Oh, that's in *our* config, :facepalm: That's a bug in `virtualenv`

well that should be fixed on the next release of `virtualenv` so we can cleanup that comment and exclude `virtualenv==20.36.0` at that time.


> More importantly, its ecosystem is a complete dog's breakfast with a multitude of tools for the same task.

I won't argue over the ecosystem. I only ever considered python to be a scripting language, not a heavy lifter like Rust or C. :shrug:

reactive-firewall:
I've added a workaround for the pip-compile side of the bug regarding `filelock` in PR #313 that, once merged, should help with that side.

Also, upon further consideration, I'm thinking `pip-compile`, un-stability, may warrant having support for other (e.g., `uv`) tooling for pinning our development groups, but the record-of-truth will continue to be the `pyproject.toml` (for all categories of direct dependencies and groups, build/runtime/dev/etc.).

Other considerations we should look at early (e.g. before we get to `v6.x`):

* there is an experimental `pip lock [-o OUTPUT_FILE] .` feature of `pip` that generates lock files (and uv handles pip stuff via `uv pip` already)
  * there is already support for --group (e.g., `pip lock -o dev-requirements.in.lock --group dev .`)
* `prek` is using `uv pip install` under the hood (might already be multi-tooled)

PR #304, opened 2026-02-15
Fix warning in tests

DimitriPapadopoulos:
```
test_piplicenses.py::TestGetLicenses::test_format_rst_default_filter
  /home/runner/work/pip-licenses/pip-licenses/test_piplicenses.py:137: DeprecationWarning: The frontend.OptionParser class will be replaced by a subclass of argparse.ArgumentParser in Docutils 2.0 or later.
    To get default settings, use frontend.get_default_settings().
    settings = docutils.frontend.OptionParser(

test_piplicenses.py: 72 warnings
  /opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/optparse.py:1000: DeprecationWarning: The frontend.Option class will be removed in Docutils 2.0 or later.
    option = self.option_class(*args, **kwargs)
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    "docutils",
+    "docutils>=0.19", # frontend.get_default_settings()
     "wheel",
     "mypy",
     "pip-tools",
--- a/test_piplicenses.py
+++ b/test_piplicenses.py
@@ -133,10 +133,10 @@ class TestGetLicenses(CommandLineTestCase):
     @staticmethod
     def check_rst(text: str) -> None:
         parser = docutils.parsers.rst.Parser()
-        components = (docutils.parsers.rst.Parser,)
-        settings = docutils.frontend.OptionParser(
-            components=components
-        ).get_default_values()
+        component = docutils.parsers.rst.Parser
+        settings = docutils.frontend.get_default_settings(
+            component  # type: ignore[arg-type]  # https://sourceforge.net/p/docutils/patches/216/
+        )
         settings.halt_level = 3
         document = docutils.utils.new_document("<rst-doc>", settings=settings)
         parser.parse(text, document)


test_piplicenses.py
```diff
@@ -133,10 +133,9 @@ def _create_license_columns(self, table, output_fields):
     @staticmethod
     def check_rst(text: str) -> None:
         parser = docutils.parsers.rst.Parser()
-        components = (docutils.parsers.rst.Parser,)
-        settings = docutils.frontend.OptionParser(
-            components=components
-        ).get_default_values()
+        settings = docutils.frontend.get_default_settings(
+            docutils.parsers.rst.Parser  # type: ignore[arg-type]  # https://sourceforge.net/p/docutils/patches/216/
+        )
```
reactive-firewall, line 139: :broom: NIT -- fixed but now slightly less readable

Consider the simpler (more focused and slightly more readable) change:

```suggestion
components = (docutils.parsers.rst.Parser,)
settings = docutils.frontend.get_default_settings(
            components,  # type: ignore[arg-type]  # https://sourceforge.net/p/docutils/patches/216/
        )
```

or if sets don't work as [the patch](https://sourceforge.net/p/docutils/patches/216/) implies

consider:
```suggestion
component = docutils.parsers.rst.Parser
settings = docutils.frontend.get_default_settings(
            component,  # type: ignore[arg-type]  # https://sourceforge.net/p/docutils/patches/216/
        )
```

Rationale: I'd prefer to keep the variable for readability here

pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
 
 [project.optional-dependencies]
 dev = [
-    "docutils",
+    "docutils>=0.19", # frontend.get_default_settings()
```
reactive-firewall, line 38: :broom: NIT -- comment and official docs don't explain why version `0.19`+

Considering:
* [RELEASE NOTES for 0.19](https://www.docutils.org/RELEASE-NOTES.html#release-0-19-2022-07-05)
* [HISTORY for 0.19](https://www.docutils.org/HISTORY.html#release-0-19-2022-07-05)

:confused: The specific min version pinned (`0.19`+) does not seem to be related to `frontend.get_default_settings()` ... is there any relevant reference documentation for this version pick, or is it just arbitrary (e.g., because it "just works")? :shrug:

No change is suggested here (even if pick is arbitrary), but I'd like to better understand the choice here.

reactive-firewall (commented):
## :+1: LGTM; Changes look fine to me as they indeed fix warnings in tests.

That said I have a question about the version requirement. I'm not sure I follow why the min version is needed. Note: to be clear I do acknowledge the security benefits of version pinning, but that does not seem to be the rationale this time.

pyproject.toml
DimitriPapadopoulos: `frontend.get_default_settings` requires docutils 0.19.

As far as I know, the docutils release notes are incorrect, `frontend.get_default_settings` was not part of docutils 0.18, it was introduced by docutils 0.19:
https://github.com/sphinx-doc/sphinx-argparse/issues/57#issuecomment-2232941654

DimitriPapadopoulos: It's indeed missing from 0.18.1:
https://sourceforge.net/p/docutils/code/HEAD/tree/tags/docutils-0.18.1/docutils/frontend.py
and can be found in 0.19:
https://sourceforge.net/p/docutils/code/HEAD/tree/tags/docutils-0.19/docutils/frontend.py
```python
Interface function:
   `get_default_settings()`.  New in 0.19.
```
```python
def get_default_settings(*components):
    """Return default runtime settings for `components`.

    Return a `frontend.Values` instance with defaults for generic Docutils
    settings and settings from the `components` (`SettingsSpec` instances).

    This corresponds to steps 1 and 2 in the `runtime settings priority`__.

    __ https://docutils.sourceforge.io/docs/api/runtime-settings.html
       #settings-priority
    """
    with warnings.catch_warnings():
        warnings.filterwarnings('ignore', category=DeprecationWarning)
        return OptionParser(components).get_default_values()
```

test_piplicenses.py
DimitriPapadopoulos: I thought sets don't work, but you're probably right:
https://sphinx-docutils.readthedocs.io/en/latest/docutils.frontend.html#docutils.frontend.get_default_settings

DimitriPapadopoulos: Actually, with a tuple it fails with:
```
E           AttributeError: 'tuple' object has no attribute 'settings_spec'
```

pyproject.toml
reactive-firewall: :bow: Thanks for the thorough explanation.

PR #127, opened 2022-10-20
Release 3.5.5

raimon49:
This release branch is fix #126 

* [x] Implement function
* [x] Testing
* [x] Add a document
* [x] Bump version
* [x] Release to PyPI

PR #305, opened 2026-02-15
Get rid of Python 3.5 left-over

DimitriPapadopoulos:
**EDIT**:

> This is just a test, not the source code itself. Tests run with Python ≥ 3.9. Attempting to protect from that is over-engineering. Besides, f-strings were introduced in Python 3.6, so any attempt to use Python 3.5 would break havoc well before reaching this test.

_Originally posted by @DimitriPapadopoulos in https://github.com/raimon49/pip-licenses/pull/305#discussion_r2810101839_

### DIFF
--- a/test_piplicenses.py
+++ b/test_piplicenses.py
@@ -614,10 +614,6 @@ class TestGetLicenses(CommandLineTestCase):
         self.assertEqual("|", table.junction_char)
         self.assertEqual(HRuleStyle.HEADER, table.hrules)
 
-    @unittest.skipIf(
-        sys.version_info < (3, 6, 0),
-        "To unsupport Python 3.5 in the near future",
-    )
     def test_format_rst_without_filter(self) -> None:
         piplicenses.importlib_metadata.distributions = (
             importlib_metadata_distributions_mocked


test_piplicenses.py
```diff
@@ -614,10 +614,6 @@ def test_format_markdown(self) -> None:
         self.assertEqual("|", table.junction_char)
         self.assertEqual(HRuleStyle.HEADER, table.hrules)
 
-    @unittest.skipIf(
-        sys.version_info < (3, 6, 0),
-        "To unsupport Python 3.5 in the near future",
-    )
```
reactive-firewall, line 620: :no_good: the skip logic is still technically correct; the test does require python 3.6+

Consider just updating the message:

```suggestion
    @unittest.skipIf(
        sys.version_info < (3, 6, 0),
        "Python 3.5 and earlier have long since been deprecated. Please update your Python.",
    )
```

Or a similar message.

Otherwise, please open a GHI ticket and document the rationale for why leaving the skip logic is reductive (e.g., slower due to pointless `sys.version` call), and replace with a short comment in code for posterity (E.g., `# needs python >= 3.6.0`).

---

Essentially something to keep the knowledge documented for posterity.

reactive-firewall (changes requested):
While I agree that python 3.5 is not even supported in theory anymore, I don't like outright removing the min-requirements silently.

test_piplicenses.py
DimitriPapadopoulos: This is just a test, not the source code itself. Tests run with Python ≥ 3.9. Attempting to protect from that is over-engineering. Besides, f-strings were introduced in Python 3.6, so any attempt to use Python 3.5 would break havoc well before reaching this test.

reactive-firewall: :thinking: Very true about the f-strings, your point is well made; We can just go with a short comment before release (e.g., just should be mentioned in the notes when released)

**Next Steps**:
 - [x] approve changes
 - [x] merge into dev
 - [x] mention in notes (:bulb: just focus on python 3.9 min requirement instead)
 - [ ] resolve this discussion

DimitriPapadopoulos: Something "applied changes that may definitively break Python < 3.9"?

DimitriPapadopoulos: Since we don't test, we don't know when pip-licences actually stopped working with Python 3.5, 3.6 or 3.8. We only know these already well-documented facts:
- pip-licenses stopped supporting Python 3.8 with 893835e / #206 / [v-5.0.0](https://github.com/raimon49/pip-licenses/releases/tag/v-5.0.0)
  > * Dropped support Python 3.8
- pip-licenses stopped supporting Python 3.7 with 129cadb / #129 / [v-4.0.0](https://github.com/raimon49/pip-licenses/releases/tag/v-4.0.0)
  > * Dropped support Python 3.7
- pip-licenses stopped supporting Python 3.6 with 2fb40bc / #127 / [v-3.5.5](https://github.com/raimon49/pip-licenses/releases/tag/v-3.5.5)
  > * Dropped support Python 3.6
- pip-licenses stopped supporting Python 3.5 with a11d0ef / #77 / [v-3.0.0](https://github.com/raimon49/pip-licenses/releases/tag/v-3.0.0)
  > * Dropped support Python 3.5

Do we really need to add more?

stefan6419846: At least with https://github.com/raimon49/pip-licenses/commit/129cadb2c565cb2294bd957cd3c3fc0d76e3a5b1, Python <= 3.7 stopped working. `importlib.metadata` is a new module since Python 3.8 and with the mentioned commit, support for using the backport if required has been removed.

reactive-firewall: @stefan6419846 thanks for the insight, as always I appreciate your input.

---

@DimitriPapadopoulos,
Regarding:
> Do we really need to add more?

:100: Nope, we good. TL;DR - that was a not to myself so I didn't miss stuff like you quoted about 893835eb41c0442db852d44df679dd03a7be88b3. Thanks for the leg work, you've saved some of my time, again.

PR #306, opened 2026-02-15
Improve package home page discovery

DimitriPapadopoulos:
Core Metadata 1.2 (PEP 753) compliance.

### DIFF
--- a/piplicenses.py
+++ b/piplicenses.py
@@ -97,22 +97,18 @@ SUMMARY_OUTPUT_FIELDS = (
 
 
 def extract_homepage(metadata: Message) -> str | None:
-    """Extracts the homepage attribute from the package metadata.
+    """Extracts the home page from the package metadata.
 
-    Not all python packages have defined a home-page attribute.
-    As a fallback, the `Project-URL` metadata can be used.
-    The python core metadata supports multiple (free text) values for
-    the `Project-URL` field that are comma separated.
+    Retrieve home page from the PEP 753 `Project-URL` metadata.
+    As a fallback, try the Core Metadata 1.0 home-page attribute.
+    If all else fails, try other PEP 753 `Project-URL` labels.
 
     Args:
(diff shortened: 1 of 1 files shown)


piplicenses.py
```diff
@@ -97,36 +97,45 @@
...
+    # https://packaging.python.org/en/latest/specifications/core-metadata/#core-metadata-project-url
+    homepage = candidates.get("homepage")
+    if homepage is not None:
+        return homepage
+
+    # fall back to deprecated Core Metadata 1.0
+    # https://packaging.python.org/en/latest/specifications/core-metadata/#home-page
+    homepage = metadata.get("home-page", None)
+    if homepage is not None:
+        return homepage
```
reactive-firewall, line 129: :+1: This is better

**Additional Comment**:

[FUTURE Considerations](https://peps.python.org/pep-0753/#future-considerations)

> This PEP does not stipulate or require any future metadata changes.
> 
> However, per [Metadata producers](https://peps.python.org/pep-0753/#metadata-producers) and [Conventions for Project-URL labels](https://peps.python.org/pep-0753/#conventions-for-labels), we identify the following potential future goals for a new major release of the core metadata standards:
> 
> Outright removal of support for Home-page and Download-URL in the next major core metadata version. If removed, package indices and consumers MUST reject metadata containing these fields when said metadata is of the new major version.
> Enforcement of label normalization. If enforced, package producers MUST emit only normalized Project-URL labels when generating distribution metadata, and package indices and consumers MUST reject distributions containing non-normalized labels. Note: requiring normalization merely restricts labels to lowercase text, and excludes whitespace and punctuation. It does NOT restrict project URLs solely to the use of “well-known” labels.
> These potential changes would be backwards incompatible, hence their inclusion only in this section. Acceptance of this PEP does NOT commit any future metadata revision to actually making these changes.

As we may need to validate and reject (as a meta-data consumer) in the future, this fall-back is especially well done.

Issue #316, opened 2026-03-09
Split single file implementation into python module in `pip-licenses/`


reactive-firewall:
For maintainability, it is time to split the code into a modular python module

Namely:

* `pip-licenses/__init__.py` - imports, constants, etc. and re-exports other components
* `pip-licenses/__main__.py` - entry point, holds `main()` - this file should also wrap anything necessary to ensure drop-in behavior of `pip-licenses/` for `pip-licenses.py`
* `pip-licenses/core.py` - core logic
* `pip-licenses/cli`**[**`/`**|**`.py`**]** - argparse, config reading, etc.
* `pip-licenses/output/` - printing, outputs, formatting, etc.
  * `pip-licenses/output/__init__.py`
  * `pip-licenses/output/csv.py`
  * `pip-licenses/output/markdown.py`
  * `pip-licenses/output/JSON.py`
  * etc.
* `pip-licenses/errors`**[**`/`**|**`.py`**]** - warnings, errors handlers, etc.
* `pip-licenses/sorting.py` - sorting utilities
* etc.

reactive-firewall:
The majority of this is implemented in the 5 beta for version 6 now.

The result is 3 key components: the core package (consisting of the core functionality and utilities), the command line interface (consisting of configuration parsing and entry-points for the command line) and output package (consisting of presettable functionality and output formatting)

The API should begin to stabilize (via copious use of `__all__`) and pave the way for work on GHI #81 in future iterations.

reactive-firewall:
The API has begun to stabilize (via copious use of __all__) with v6.0.0b8, and should start to unblock the rest of the work on the version 6 series.

reactive-firewall:
This is mostly complete and expected to be part of the next major release now (e.g., when it will be officially closed).

Issue #334, opened 2026-05-26
Chore: look into idea of sorting options for multiple matching results on a single package (namely the license files)


reactive-firewall:
Task: research what it would take to implement/maintain fine-grain sorting of multiple matches from work on #242 through user config/CLI flags

reactive-firewall:
Proposed plan for sorting (*W.I.P.*):

By default:

1. If package metadata provides multiple `License-File`s  then return them in the order they are provided by internal metadata (referred to as 'as specified' by the maintainer)
2. If additional notice files are to be returned, unless already listed from (1.) are to be appended _after_ official metadata files
3. If additional "other" files are to be returned, unless already listed from (2.) are to be appended _after_ official metadata files AND _after_ any notice files, sorted by Python's `sort()` function

---

other sort orders to explore:

* `reverse` same as default but then reversed final order

* `relevance`
  1. If package metadata provides multiple `License-File`s  then return them in order by:
     A. UPPERCASE before lowercase
     B. LICENSE (case-insensitive) first if found
     C. for files with an extension/suffix (when prefix is otherwise equal):
        i. dash `-` before dot `.`
        ii. `.txt` (case-insensitive, see A.) before dot any other dot-suffix (extension) e.g. `.md`, `.rst`, `.rst.txt` etc.
        iii. `.md` (case-insensitive, see A.) before dot any other non-`.txt` dot-suffix (extension) e.g. `.txt.md`, `.markdown`, `.rst`, `.rst.txt` etc.
        iv. single dot extensions before multi-dot extensions
        v. by extension/suffix length
        vi. finally by internal `sort()`
     D. LICENSE.* (case-insensitive, with an extension/suffix) by case and name length from shortest to longest
     E. COPYRIGHT* (case-insensitive, with an optional extension/suffix) by case and name length from shortest to longest
     F. COPY* (case-insensitive, with an optional extension/suffix) by case and name length from shortest to longest
     G. NOTICE* (case-insensitive, with an optional optional extension/suffix) by case and name length from shortest to longest
     H. anything else by case, `sort()`, and name length from shortest to longest
  2. If additional "other" files are to be returned, unless already listed from (1.) are to be integrated with (by being inserted just after) official metadata files AND  sorted by same rules A-D as (1.) and an additional rule: D.2 official listed files before others, AND the remaining rules E-H from (1.)
  3. If additional notice files are to be returned, unless already listed from (2.) are to be appended _after_ everything else following the same rules as (2.)

[...]

PR #346, opened 2026-07-15
Patch multi-licenses workaround #242

### LINKED ISSUE #71: Packages with multiple LICENSES files not shown in --with-license-file
```bash
$ pip install -U pip pip-licenses
$ pip install opencv-python-headless==4.2.0.34
$ pip-licenses --from=mixed --with-license-report
```

The `opencv-python-headless` wheels include two license files, `LICENSE.txt` and `LICENSE-3RD-PARTY.txt`, but only `LICENSE-3RD-PARTY.txt` is displayed. Could support be added to display both license files since some packages like `opencv` have multiple files covering different parts of the library.

### LINKED ISSUE #242: Fix/Improve handling of multiple license files per package in ALL formats
Goal: Define currently undefined behavior of `--with-license-file` when encountering multiple license files.

Also add an explicit test-case to ensure stability in such scenarios.

### LINKED ISSUE #309: Refactor legacy license file search logic too
Goal: move legacy logic of searching for license files to a fallback role, and expand to check all paths in distribution `files`

* limited scope to arbitrary set of 'well-known' filenames like `LICENSE`
* best-effort for undeclared (e.g., lack of `License-File` metadata) license files
* test case example: ~`pytest-cache` version `1.0`~ **EDIT**: a hermetic test fixture

**Additional Comments**:

From discussion over in #30

> However, even if passed the `--with-license-file`, the output would still not have included the **undeclared** `LICENSE` file.
> 
> Might not be a bad idea to also followup on that idea during GHI #242. I'll open a sub-issue about undeclared licenses with arbitrary 'well-known' filenames like `LICENSE`/`COPYING` (the `SOURCES` file in the `.egg` did mention the `LICENSE` file so should show up in metadata `files` list)

Relevant "legacy logic":

https://github.com/raimon49/pip-licenses/blob/b1a00ac36bdb8a6267e7e35dc12a09abb9773321/piplicenses.py#L348-L356

**Test Cases specification**

With additional insight from discussions below, I've decided to just use the test-fixture idea.

1. Consider the following naïve snippet:

`setup.py`:
```python
from setuptools import setup

setup(
    name='example',
    version='1.0.0',
    packages=['example'],
    package_data={'example': ['LICENSE']},  # Include LICENSE file
    include_package_data=True,
    # No License-File entry
)
```

2. and the no setup.py test-case:

just `setup.cfg`:
```ini
[metadata]
name = example
version = 1.0.0
description = An example package

[options]
packages = find:
include_package_data = true

```

Technically, If you do not specify a MANIFEST.in file, setuptools has default behaviors for including certain types of files (like Python files, etc.), but it does not guarantee the inclusion of arbitrary files, such as a LICENSE file.

so let's add the MAINFEST.in file

```
include LICENSE
```

### LINKED ISSUE #330: dist-info data ignored in favour of file found in the package
Between pip-licenses 5.5.0 and 5.5.5, the licence output from the SasView wheel changed; I suspect this is due to now finding lowercase filenames (#208) accidentally meaning that other files were found.

SasView wheel correctly declares:

```
License: BSD-3-Clause
License-File: LICENSE.TXT
```
and `sasview-[version].dist-info/licenses/LICENSE.TXT` exists, with v5.5.0 returning that file:

```
$ pip-licenses -v
pip-licenses 5.5.0

$ pip-licenses -p sasview --with-license-file --format=json
[
  {
    "License": "BSD License",
    "LicenseFile": ".../.venv/lib/python3.13/site-packages/sasview-[VERSION].dist-info/licenses/LICENSE.TXT",
    "LicenseText": "Copyright (c) 2009-2025, ...
```

However, the sasview wheel also includes the file `sas/docs/_sources/user/license.rst.txt` (sasview ships its documentation inside the wheel as it needs to be displayed at runtime and while offline). That's an rst file with other text in, not just the licence text. That's not the version we want pip-licenses to return, but suddenly v5.5.5 has started doing so:

```
$ pip-licenses -v
pip-licenses 5.5.5

$ pip-licenses -p sasview --with-license-file --format=json
[
  {
    "License": "BSD License",
    "LicenseFile": ".../.venv/lib/python3.13/site-packages/sas/docs/_sources/user/license.rst.txt",
    "LicenseText": ".. _license:\n\n...
```

My expectation (and how pip-licenses used to behave) is that pip-licenses would give me the licence file defined in the wheel metadata, not some other file that happened to be in the wheel which could be entirely unrelated data, or as in this case, the correct data but in an unwanted rst format.

reactive-firewall:
# Patch Notes

Pre-GHI #316 work to mitigate GHI #242 (by introducing `--with-license-files` as workaround)

## BREAKING CHANGES

> [!IMPORTANT]
> Moving forward all single file behavior is deprecated; Starting with `v6.1` support for single license behavior will start shifting to always supporting multi licenses. The rationale is simple: `pip-licenses` does not intend to provide legal determinations and thus _should always_ (while still allowing `--from` filtering, etc.) provide the more accurate state of multiple licenses when declared by a package.

> [!IMPORTANT]
> Due to fix in support of multiple license sources (metadata + file(s)) allow/fail logic differentiates between `BSD License`, `BSD-2-Clause` and `BSD-3-Clause` (where legacy classifier based logic may have previously treated both as "BSD License").

### New CLI flags introduced in a dedicated `license_file_options` group:

`--with-license-files`: Multi-file variant (currently: JSON, PLAIN_VERTICAL formats only). Will eventually replace `--with-license-file`.
`--no-file-paths`: Suppresses _all_ file path outputs
`--with-notice-files`: Multiple notice files. (will eventually replace deprecated `--with-notice-file`)
`--with-other-files`: Enables file discovery for unlisted: Authors, Copying, Legal files.
Validation in `_verify_args()` ensures these flags are only used with compatible combinations.

## Impacted GHIs:

 * Contributes to work on #71
 * Contributes to work on #242
 - [x] Resolves #309
 * Opens #347
 
## Included & Superseded PR/MRs

 * Includes & Supersedes #325
 * Includes & Supersedes #341
 * Includes & Supersedes #343
 * Derived from and Supersedes stefan6419846/pip-licenses-cli#132 by [Stefan](https://github.com/stefan6419846) under [MIT License](https://github.com/stefan6419846/pip-licenses-cli/blob/ef14b50a23d36e8aa53f41ec49c2ea4974373a07/LICENSE.txt)
 
 ---


.github/workflows/CD-PyPi.yml
reactive-firewall, line 1: Regarding `.github/workflows/CD-PyPi.yml`; These are all version update changes for release automation. (included in this PR to resolve minor conflict with `dev-v6` branch)

.github/workflows/python-package.yml
reactive-firewall, line 1: Regarding `.github/workflows/python-package.yml`; These are also all version update changes for CI/CD testing automation. (included to resolve another minor conflict with `dev-v6` branch)

piplicenses.py
```diff
@@ -135,12 +255,18 @@ def extract_homepage(metadata: Message) -> str | None:
         "bug tracker",
     ):
         if priority_key in candidates:
-            return candidates[priority_key]
+            return _help_get_first_of_many(candidates[priority_key])
```
reactive-firewall: These will likely need sorting support as part of GHI #334. but by selecting the first value for now should not cause a regression.

piplicenses.py
```diff
@@ -135,12 +255,18 @@ def extract_homepage(metadata: Message) -> str | None:
...
 
     return None
 
 
-PATTERN_DELIMITER = re.compile(r"[-_.]+")
+def extract_license_from_classifiers(metadata: Message) -> list[str]:
```
reactive-firewall, line 276: NIT :broom: This new utility might need to have a unit-test as part of GHI #316.

piplicenses.py
```diff
@@ -351,55 +538,169 @@ def get_pkg_included_file(
...
+                leg_value
+                and (leg_value is not None)
+                and (len(leg_value) > 0)
+                and (leg_value is not LICENSE_UNKNOWN)
+            )
         }
+
+        for filtered_key in filtered_legacy_info:
+            pkg_info[filtered_key] = filtered_legacy_info[filtered_key]
+
```
reactive-firewall, line 643: This is partly to improve modularity in advance of GHI #316 while working on GHI #242

piplicenses.py
```diff
@@ -351,55 +538,169 @@ def get_pkg_included_file(
...
         metadata = pkg.metadata
         for field_name, field_selector_fns in METADATA_KEYS.items():
             value = None
             for field_selector_fn in field_selector_fns:
                 # Type hint of `Distribution.metadata` states `PackageMetadata`
-                # but it's actually of type `email.Message`
+                # but it's actually of type `email.message.Message`
```
reactive-firewall, line 649: Rationale: while the code imports this as `email.Message` the documentation is looked up at the canonical `email.message.Message`

reactive-firewall:
RFC: @stefan6419846

As this work was inspired by your work @stefan6419846, I humbly ask if you could take a brief look-over the changes and provide any feedback you might have (even just a thumbs up if nothing jumps out). I appreciate the time and efforts you made on this issue, both on your own project and regarding this project.

RFC: @johnthagen

As this work is aimed at resolving GHI #71 raised by @johnthagen I also request any comments or feedback you may have regarding this PR.

Again, thank you both for your time on this matter. :bow:

piplicenses.py
```diff
@@ -93,6 +94,89 @@
...
+        >>> metadata = DummyMessage([
+        ...     "Source,   ",
+        ... ])
+        >>> extract_urls(metadata)
+        {'source': None}
+    """
+    _urls: dict[str, str | list[str] | None] = {}
+    for entry in metadata.get_all("Project-URL", []):
+        key, value = entry.split(",", 1)
+        if key.strip().lower() in _urls:
```
stefan6419846: Repeating `key.strip().lower()` all over (similar things apply to `value.strip()`) is just an overhead and harms readability.

I suggest introducing dedicated variables - either through the Walrus operator or in a dedicated definition line.

piplicenses.py
```diff
@@ -93,6 +94,89 @@
...
+            if not isinstance(_urls[key.strip().lower()], list):
+                # MyPy is a bit lost by this point, but we can only ever reach this point iff:
```
stefan6419846: Why is MyPy lost here? Is there anything which could avoid this, like using a dedicated variable for the retrieved non-list URL or similar?

piplicenses.py
```diff
@@ -93,6 +94,89 @@
 }
 
 
+def extract_urls(metadata: Message) -> dict[str, str | list[str] | None]:
+    """Extract normalized project URLs from message metadata.
+
+    This scans all ``Project-URL`` entries in the given metadata object and
+    builds a dictionary keyed by the lowercased, stripped project name.
```
stefan6419846: "Stripped project name" sounds wrong here. Isn't this the URL label?

piplicenses.py
```diff
@@ -351,55 +550,169 @@ def get_pkg_included_file(
...
+        if args.with_license_files:  # conditional for < v6+
+            pkg_texts: list[str | None] | None = (
+                get_pkg_license_texts_from_disk(
+                    pkg,
+                    filelist=cast(list[str], pkg_info["license_files"])
+                    if pkg_info["license_files"]
+                    else None,
+                )
+            )
+            if pkg_texts and pkg_texts is not None:
```
stefan6419846: Isn't the second part already covered by the first one?

piplicenses.py
```diff
@@ -351,55 +550,169 @@ def get_pkg_included_file(
...
+                pkg_info["license_texts"] = cast(list[str], pkg_texts)
 
-        classifiers: list[str] = metadata.get_all("classifier", [])
-        pkg_info["license_classifier"] = find_license_from_classifier(
-            classifiers
-        )
+        if args.with_other_files:  # conditional for < v6+
+            OTHER_FILES_PATTERN = r"[Aa][Uu][Tt][Hh][Oo][Rr][Ss].*|[Cc][Oo][Pp][Yy][Ii][Nn][Gg].*|[Ll][Ee][Gg][Aa][Ll].*"
```
stefan6419846, line 671: Isn't COPYING already included in the license files? To be honest, I do not remember every seeing a file named LEGAL. Additionally, it will always miss non-standard names which should be reported in theory.

piplicenses.py
```diff
@@ -351,55 +550,169 @@ def get_pkg_included_file(
...
+                    return item.encode(
+                        args.filter_code_page, errors="ignore"
+                    ).decode(args.filter_code_page)
+                except AttributeError as _cause:  # pragma: no cover
+                    _cntx_dtls = f"{item} can not be safely filtered with {args.filter_code_page}"
+                    if not isinstance(item, str):
+                        _context_details = (
+                            f"{type(item)} can not be filtered as a string"
+                        )
+                    raise ValueError(_cntx_dtls) from _cause
```
stefan6419846: `_cntx_dtls` or `_context_details` should be used consistently.

piplicenses.py
```diff
@@ -526,17 +889,52 @@ def create_licenses_table(
                     cast(str, pkg["license"]),
                     cast(str, pkg["license_expression"]),
                 )
-                license_str = "; ".join(sorted(license_set))
+                _sorted_license_set = (
+                    sorted(license_set) if license_set else []
+                )
+                _normalized_lisense_set = {
```
stefan6419846: Typo: `lisense` -> `license`

piplicenses.py
```diff
@@ -711,11 +1109,26 @@ class PlainVerticalTable(PrettyTable):
     def get_string(self, **kwargs: str | list[str]) -> str:
         options = self._get_options(kwargs)
         rows = self._get_rows(options)
+        show_paths = "LicenseFiles" in kwargs["fields"]
 
         output = ""
         for row in rows:
-            for v in row:
-                output += f"{v}\n"
+            index = 0
```
stefan6419846, line 1120: This seems to mostly be a verbose copy from my code with some small formatting changes. Please make sure to include proper credits where required, id est where you created copies of non-trivial code.

test_piplicenses.py
```diff
@@ -864,6 +1048,280 @@ def test_case_insensitive_partial_match_set_intersect(self) -> None:
...
+        self.assertIsNotNone(
+            _handcrafted_metadata, "Can't create metadata at all"
+        )
+        _handcrafted_metadata["Project-URL"] = _url_key_comma_value_1
+        self.assertIsNotNone(
+            _handcrafted_metadata.get("Project-URL", None),
+            "Can't read metadata at all",
+        )
+        # Note:
+        # mypy is brain dead and thinks strings can't be in things:
```
stefan6419846: This somehow sounds wrong and if it really can be verified, it should most likely be reported to upstream.

stefan6419846 (commented):
* I personally would avoid doing the dependency updates and unrelated changes in this PR to keep the diff as small as possible, but this is a matter of taste.
* Where are the new options documented?

piplicenses.py
reactive-firewall: :+1: I concur, `key.strip().lower()` should be much more D.R.Y.

I'm considering something like this:
```suggestion
        _norm_key: str = key.strip().lower()
        if _norm_key in _urls:
```

and similar elsewhere...

My rationale: normalize key to lowercase as a dedicated def allows refactor to a filter function in future if validation is ever extended (with less over-head for the one var per iteration than a bunch of function calls over-and-over)

piplicenses.py
reactive-firewall: > Why is MyPy lost here?

:shrug: not quite sure

> Is there anything which could avoid this, like using a dedicated variable for the retrieved non-list URL or similar?

worth looking into again, but the issue had to do with MyPy complaining that list[list[str]] (nested lists) were a problem here, (not accounting for the defensive check that we are not dealing with a list)

I suspect a dedicated variable (while negligibly less efficient at extreme scales) could work and thus is worth reconsidering the trade-off of syntactic-complexity and cognitive-complexity here

piplicenses.py
reactive-firewall: Yup, that's a semantic error!

This should instead be:

```suggestion
    builds a dictionary keyed by the lowercased, stripped, label value of
    the pair (e.g., 'source' for 'Source, https://github.com/user/repo.git').
```

piplicenses.py
reactive-firewall: yeah, that looks odd, I think this is could be refactored to (equally defensive style):

```suggestion
            if pkg_texts and pkg_texts is not [None]:
```

I also recall MyPy complained about:

```python
            if pkg_texts:
                pkg_info["license_texts"] = cast(list[str], pkg_texts)
```

because `pkg_texts` needs to be non-`None` to cast ... I'll test this later and either add a comment or refactor

piplicenses.py
reactive-firewall: While, I have seen projects with the filename "LEGAL" instead of "LICENSE" (e.g., where a licensed patent was being attributed, separate from the license) but _thinking back none of those were Python_ projects, so I did a quick search:
https://github.com/search?q=path%3ALEGAL+language%3APython&type=Repositories&ref=advsearch&l=Python

and don't see any hosted on GitHub :shrug: using the filename `LEGAL`, and these sorts of details are typically in the pythonic `NOTICE` files instead.

So I concur, that pattern is not really needed here (albeit harmless).

---

but regarding duplication of `COPYING|copying` file names, that is a bit of technical debt, as the plan is to move to listing metadata declared (by the package authors themself) `License-Files` as the content of the resulting table's `License File(s)` and discovered, but unlisted/deprecated, well-known named files (e.g., case-insensitive `LICENSE{.*}|COPYING{.*}|AUTHORS{.*}|CONTRIBUTORS{.*}` etc.) in separate, and _optional_ fields, to re-enforce that these are _not attested_ licenses from the package authors/maintainers, I'm moving them over to `--with-other-files` (and yeah, the missing documentation of that is clearly needed). So this is indeed where filtered paths belong (albeit, as you say, the heuristic used here and in legacy implementations will always miss any "non-standard names which should be reported in theory", and that detail should also be explained in some documentation)

---

I've opened GHI #349 to track the work on new documentation

piplicenses.py
reactive-firewall: > `_cntx_dtls` or `_context_details` should be used consistently.

Considering something more like this:
```suggestion
                    _context_details = (
                        f"{item} can not be safely filtered with {args.filter_code_page}"
                    )
                    if not isinstance(item, str):
                        _context_details = (
                            f"{type(item)} can not be filtered as a string"
                        )
                    raise ValueError(_context_details) from _cause
```

piplicenses.py
reactive-firewall:
```suggestion
                _normalized_license_set = {
```

and similar elsewhere

test_piplicenses.py
reactive-firewall: With version `1.19.1` it is reproducible on my system, but can be ignored (as done around line 1115)
```suggestion
        # mypy (v1.19.1 for Python v3.9) is brain dead and thinks strings can't be in things:
```

mypy (namely version 1.19.1) also errored/complained about my first iteration (which I naïvely just believed it when it complained):
```python
# in place-of the self.assertTrue(...) expression
self.assertIn(
    _url_value_1, _handcrafted_metadata.get("Project-URL", None),
    "Can't find {_url_value_1} in metadata after setting it",
)
```

I'm not interested in doing the leg-work to report it now that mypy has dropped support of python 3.9 in recent v2+ versions (where I imagine they may have improved it :shrug:). The `# type: ignore[operator]` allows str-types to use the `in` operator anyway.

piplicenses.py
reactive-firewall: Update: after looking into this a bit more, the issue was more nuanced; and the variable workaround is indeed a "trade-off of syntactic-complexity and cognitive-complexity here", so I'm just going to clarify the vague comments instead of bending-over backwards for MyPy.

TL;DR - walk through for posterity

```python
def extract_urls(metadata: Message) -> dict[str, str | list[str] | None]:
    """Extract normalized project URLs from message metadata.
       ...OMITTED FOR BREVITY...
    """
    _urls: dict[str, str | list[str] | None] = {}
    for entry in metadata.get_all("Project-URL", []):
        key, value = entry.split(",", 1)
        _norm_key: str = key.strip().lower()
        if _norm_key in _urls:
            if not isinstance(_urls[_norm_key], list):
                # MyPy is a bit lost by this point, but we can only ever reach this point iff:
                # 1. _urls[_norm_key] exists (so type can no longer be assumed to be None)
                # 2. _urls[_norm_key] is truthy
                # 3. _urls[_norm_key] is NOT of type list (so combined with axiom 1 gives str-type)
                # ... are ALL true (otherwise does nothing).
                # So to help MyPy follow along we MUST store the value in a transient variable here
                _norm_list: list[str] = [
                    cast(str, _urls[_norm_key]),
                ]
                # so MyPy realizes _urls[_norm_key] is now of type list[str]
                _urls[_norm_key] = _norm_list
                # Otherwise we could have just done this:
                # _urls[_norm_key] = [
                #     _urls[_norm_key],  # type: ignore[list-item]  # ty: ignore[unused-type-ignore-comment]
                # ]
            # We can only ever reach this point iff:
            # 1. _urls[_norm_key] exists (so type can no longer be assumed to be None)
            # 2. _urls[_norm_key] is truthy
            # 3. _urls[_norm_key] IS of type list[str] (either already or because we wrapped it)
            # But ONLY because we hand-held MyPy via unnecessary '_norm_list' does it catch on.
            if value.strip() not in set(
                cast(list[str], _urls[_norm_key])
            ):  # deduplicate and merge by key
                # in the case value was ALREADY a list but NOT merged in _urls[_norm_key]
                # yet MyPy still thinks _urls[_norm_key] might not be a list[str]
                # and thus complains it has no append attribute
                # so more hand-holding:
                _swap_val_list: list[str] = cast(list[str], _urls[_norm_key])
                _swap_val_list.append(value.strip())
                _urls[_norm_key] = _swap_val_list
                # Otherwise we could have just done this:
                # _urls[_norm_key].append(value.strip())  # type: ignore[union-attr]  # ty: ignore[unused-type-ignore-comment]
        else:
            _urls[_norm_key] = value.strip() or None
    return _urls

```

Ommiting the comments this is a choice between:

```python
def extract_urls(metadata: Message) -> dict[str, str | list[str] | None]:
    _urls: dict[str, str | list[str] | None] = {}
    for entry in metadata.get_all("Project-URL", []):
        key, value = entry.split(",", 1)
        _norm_key: str = key.strip().lower()
        if _norm_key in _urls:
                _norm_list: list[str] = [
                    cast(str, _urls[_norm_key]),
                ]
                _urls[_norm_key] = _norm_list
            if value.strip() not in set(
                cast(list[str], _urls[_norm_key])
            ):
                _swap_val_list: list[str] = cast(list[str], _urls[_norm_key])
                _swap_val_list.append(value.strip())
                _urls[_norm_key] = _swap_val_list
        else:
            _urls[_norm_key] = value.strip() or None
    return _urls

```

**OR**

```python
def extract_urls(metadata: Message) -> dict[str, str | list[str] | None]:
    _urls: dict[str, str | list[str] | None] = {}
    for entry in metadata.get_all("Project-URL", []):
        key, value = entry.split(",", 1)
        _norm_key: str = key.strip().lower()
        if _norm_key in _urls:
                _urls[_norm_key] = [
                     _urls[_norm_key],  # type: ignore[list-item]  # ty: ignore[unused-type-ignore-comment]
                ]
            if value.strip() not in set(
                cast(list[str], _urls[_norm_key])
            ):
                _urls[_norm_key].append(value.strip())  # type: ignore[union-attr]  # ty: ignore[unused-type-ignore-comment]
        else:
            _urls[_norm_key] = value.strip() or None
    return _urls

```

So I see the second option as more readable (also note that the `_swap_val_list: list[str] = cast(list[str], _urls[_norm_key])` could be moved out of the condition and used in the condition to DRY up that strategy, but then it becomes clear that we already did a cast in the case of a non-list, so `_norm_key` and `_swap_val_list` are redundant in that execution path)

So with all that I think it is a hasty abstraction to use variables just to satisfy a type linter (e.g., MyPy) I'll instead just reference this discussion as a single comment.

---

```suggestion
                # MyPy is a bit lost by this point, (See Discussion in PR #346)
                # <link to this comment>
```

piplicenses.py
reactive-firewall: @stefan6419846 

:bow: I am happy to attribute as much as [that whole commit](https://github.com/raimon49/pip-licenses/commit/53e471222c7db1ee90c33cf7efd8bfc2d764b921) to you as co-author if you would prefer. (E.g., I can roll-back, re-attribute, graph the other changes back via interactive re-base, and force push back to this branch if you prefer the `git` history to reflect your authorship)


Please advise;
1. How would you prefer to be credited,
2. and clarify if the resulting 539 LoC are (or not) intended to be considered "substantial portions of the " _[pip-licenses-cli's]_ "Software" as per [MIT License](https://github.com/stefan6419846/pip-licenses-cli/blob/ef14b50a23d36e8aa53f41ec49c2ea4974373a07/LICENSE.txt)


---
# TL;DR - Clarification on the serious topic of attribution raised here.

Indeed this workaround is partially based on the related https://github.com/stefan6419846/pip-licenses-cli/pull/132/changes (mentioned in #71 as a fix for @johnthagen) from your project [pip-licenses](https://github.com/stefan6419846/pip-licenses-cli) release (circa [PR 132](https://github.com/stefan6419846/pip-licenses-cli/pull/132), and also at the time of this comment under an MIT license, granting explicit rights to copy, modify, distribute and publish, (and others omitted here for brevity), under the condition _substantial_ portions retain the copyright notice, [full license text](https://github.com/stefan6419846/pip-licenses-cli/blob/ef14b50a23d36e8aa53f41ec49c2ea4974373a07/LICENSE.txt))

Relevant Commits:
See [stefan6419846/pip-licenses-cli](https://github.com/stefan6419846/pip-licenses-cli/pull/132) project's [PR 132](https://github.com/stefan6419846/pip-licenses-cli/pull/132)
 1. Less than 309 LoC from [0b1c2ae56d0d8bd0bac078bf758cc770da5527b0](https://github.com/stefan6419846/pip-licenses-cli/commit/0b1c2ae56d0d8bd0bac078bf758cc770da5527b0)
 2. Less than 337 LoC from [a9ee12e83cf6b149dbee0f548d479187ad55fe46](https://github.com/stefan6419846/pip-licenses-cli/pull/132/changes/a9ee12e83cf6b149dbee0f548d479187ad55fe46)

Derived & modified as part of [This PR](https://github.com/raimon49/pip-licenses/pull/346):
 1. Less than 539 LoC in [53e471222c7db1ee90c33cf7efd8bfc2d764b921](https://github.com/raimon49/pip-licenses/commit/53e471222c7db1ee90c33cf7efd8bfc2d764b921)

My rationale was:

1. Selective copy of the fix (no intention of misrepresentation - see #71)
2. Much of the code was refactored and so tagging you as co-authoring the commit seemed presumptuous at the time.

> [!IMPORTANT]
> I apologise to anyone who was confused in regards to [This PR](https://github.com/raimon49/pip-licenses/pull/346) being derived from [PR 132](https://github.com/stefan6419846/pip-licenses-cli/pull/132) and thus code by [Stefan](https://github.com/stefan6419846) under [MIT License](https://github.com/stefan6419846/pip-licenses-cli/blob/ef14b50a23d36e8aa53f41ec49c2ea4974373a07/LICENSE.txt)

---

stefan6419846: I do not think that having a large discussion here about attribution makes sense and when the obligations are triggered. It is MIT-licensed code and I am completely fine in this case if a short notice about the inspiration/origin is part of the relevant commit message.

viktoriavikulya (commented):
Tested locally and it works. Consider extracting the retry logic into a shared utility — we have similar code in the auth module.

piplicenses.py
reactive-firewall: > I am completely fine in this case if a short notice about the inspiration/origin is part of the relevant commit message.

Commit message has been added in ec4d04c3500a961696c382988d075c145d2cdc9e (re-wrote the git history for 53e471222c7db1ee90c33cf7efd8bfc2d764b921, to ensure to mention you in the relevant commit message(s) as requested)

Again, apologies for the oversight, and a big thank you, for your time and effort on this issue!

johnthagen:
@reactive-firewall Does this fix:

- #71

I see mentions of "contributes" to, so I was curious what parts are still left pending, if any.

And is the plan for this to be included in `pip-licenses` 6?

- #352

reactive-firewall:
> @reactive-firewall Does this fix:
> 
> * [Packages with multiple LICENSES files not shown in --with-license-file #71](https://github.com/raimon49/pip-licenses/issues/71)
> 
> I see mentions of "contributes" to, so I was curious what parts are still left pending, if any.

Not fully, no (I'd call this a _workaround_, **not** a fix), this does improve the situation; however only for some formats, namely `--format=json` and `--format=plain-vertical`. Notably HTML and CSV formats are unresolved / not implemented yet (and of-course the rest of GHI #242 enhancements are un-implemented still).

There is also a bunch of technical debt around documentation and functional testing for #71.

This workaround is something that I've had `git stash`ed in my local repo for almost a year now  (from September 2025), and needed to be merged before the WIP on #316 could be litmus tested) and fixes #309 in such a way that can be migrated.

> And is the plan for this to be included in `pip-licenses` 6?

Kinda. This is a transitional step included in the Major updates [planned for the `pip-licenses` 6 series](https://github.com/raimon49/pip-licenses/milestone/3). Although I categorized this as part of [Initial Compliance with Python Standards](https://github.com/raimon49/pip-licenses/milestone/5).

Regarding
> * [Release v6.0.0 (Alpha release) preparation #352](https://github.com/raimon49/pip-licenses/issues/352)

This is not set in stone yet. Honestly, the biggest factor is: I'll really don't care to maintain this fix (as-is) long-term, not until the maintainability improvements for #316 have landed (but I also didn't want regressions in the middle of that work).

The main goal of #352 is to land some documentation improvements that were specifically requested there; That will _only_ be a pre-release for those interested in the latest cutting-edge.

hopefully that helps answers your questions @johnthagen

reactive-firewall:
> [!CAUTION]
> # PR Superseded
> This transitional patch has now been mostly (except for `test_piplicenses.py`) superseded by work in #361 (circa development [Beta `6.0.0b7`](https://github.com/raimon49/pip-licenses/tree/v-6.0.0b7))

